PR #83, opened 2024-05-21
75 create open api endpoint to search for cves and retrieve all information we are currently storing in our vulnerability database

Refoo0:


internal/core/vulndb/vulndb_controller.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		risk = cvss.Score()
+
+	}
+
+	return risk
+}
+
+func envHandle(env core.Environmental) core.Environmental {
```
timbastin: I think this logic should be done in the context_utils.go file. Otherwise, one is responsible for calling `GetEnvironmental` and `envHandle` afterward. If one does forget to do so, the core.Environmental struct is basically in an invalid state.

Refoo0:
@timbastin 
Resolve pull request comments

seb-kw (approved):

PR #86, opened 2024-05-23
77 implement metrics endpoint

Refoo0:


internal/core/org/org_controller.go
```diff
@@ -166,3 +168,13 @@ func (o *httpController) List(c core.Context) error {
 
 	return c.JSON(200, organizations)
 }
+
+func (o *httpController) Metrics(c core.Context) error {
+	fmt.Println("metrics")
```
timbastin: this can be removed

internal/database/repositories/org_repository.go
```diff
@@ -41,3 +41,9 @@ func (g *orgRepository) ReadBySlug(slug string) (models.Org, error) {
 	err := g.db.Where("slug = ?", slug).First(&t).Error
 	return t, err
 }
+
+func (g *orgRepository) GetSlug(orgID string) (models.Org, error) {
```
timbastin: this does not make too much sense. The function is called GetSlug and returns a whole model. I think you can just use the "Read" method

timbastin (approved):

PR #91, opened 2024-06-03
88 add the possibility to handle identified flaws

Refoo0:


internal/core/flaw/flaw_controller.go
```diff
@@ -95,22 +108,58 @@ func (c flawHttpController) Read(ctx core.Context) error {
...
+	if err != nil {
+		return echo.NewHTTPError(400, "invalid payload").WithInternal(err)
+	}
+
+	statusType := status.StatusType
+	justification := status.Justification
+
+	fmt.Println("CreateEvent", statusType, justification)
+
+	err = c.flawService.UpdateFlawState(nil, userID, flaw, statusType, &justification)
```
timbastin: I like, that we have a single transaction for the database operations of saving the flaw and creating a new event. This makes sure, that the database stays consistent. I would prefer, that we use the feature right here, by creating a new transaction. Passing nil actually disables the transaction feature.

```suggestion
        err = c.flawRepository.Transaction(func(tx core.DB) error {
           return c.flawService.UpdateFlawState(tx, userID, flaw, statusType, &justification)
        })
```

cmd/flawfix/api/api.go
```diff
@@ -315,6 +316,8 @@ func Start(db core.DB) {
 	flawRouter.GET("/", flawController.ListPaged)
 	flawRouter.GET("/:flawId/", flawController.Read)
 
+	flawRouter.POST("/:flawId/", flawController.CreateEvent)
```
timbastin: We need to make sure, that the user is allowed to update the asset. We can do that easily by using the projectScopedRBAC:
```suggestion
	flawRouter.POST("/:flawId/", flawController.CreateEvent, projectScopedRBAC("asset", accesscontrol.ActionUpdate))
```

internal/core/flaw/flaw_controller.go
```diff
@@ -95,22 +108,58 @@ func (c flawHttpController) Read(ctx core.Context) error {
...
+	var status FlawStatus
+	err = json.NewDecoder(ctx.Request().Body).Decode(&status)
+	if err != nil {
+		return echo.NewHTTPError(400, "invalid payload").WithInternal(err)
+	}
+
+	statusType := status.StatusType
+	justification := status.Justification
+
+	fmt.Println("CreateEvent", statusType, justification)
```
timbastin:
```suggestion
```

internal/core/flaw/flaw_service.go
```diff
@@ -83,3 +83,22 @@ func (s *service) UserDetectedFlaws(tx core.DB, userID string, flaws []models.Fl
...
+		Type:          models.FlawEventType(statusType),
+		FlawID:        flaw.CalculateHash(),
+		UserID:        userID,
+		Justification: justification,
+	}
+	// apply the event on the flaw
+	flaw = ev.Apply(flaw)
+
+	// run the updates in the transaction to keep a valid state
+	err := s.flawRepository.SaveBatch(tx, []models.Flaw{flaw})
```
timbastin:
```suggestion
	err := s.flawRepository.Save(tx, []models.Flaw{flaw})
```

internal/core/flaw/flaw_service.go
```diff
@@ -83,3 +83,22 @@ func (s *service) UserDetectedFlaws(tx core.DB, userID string, flaws []models.Fl
...
+	if err != nil {
+		return err
+	}
+	return s.flawEventRepository.SaveBatch(tx, []models.FlawEvent{ev})
```
timbastin:
```suggestion
	return s.flawEventRepository.Save(tx, []models.FlawEvent{ev})
```

internal/database/models/flawevent_model.go
```diff
@@ -51,3 +60,39 @@ func NewDetectedEvent(flawID string, userID string) FlawEvent {
 		UserID: userID,
 	}
 }
+
```
timbastin, line 63: if you are not using the constructor functions, you can delete those 😄

internal/core/flaw/flaw_controller.go
```diff
@@ -95,22 +108,58 @@ func (c flawHttpController) Read(ctx core.Context) error {
...
+	}
+	userID := core.GetSession(ctx).GetUserID()
+
+	var status FlawStatus
+	err = json.NewDecoder(ctx.Request().Body).Decode(&status)
+	if err != nil {
+		return echo.NewHTTPError(400, "invalid payload").WithInternal(err)
+	}
+
+	statusType := status.StatusType
```
timbastin, line 151: you need to validate, if the statusType is inside the "allowed" status types. The user can send any string as status type, but we should only accept a small subset of that like: "accepted", "markedForMitigation" ...

timbastin (approved):

PR #93, opened 2024-06-04
Improves set up for local start and improves readme

seb-kw:


.github/workflows/postgresql.yaml
```diff
@@ -2,8 +2,11 @@ name: Create and publish the custom postgresql Docker image
 
 on:
   push:
+    branches:
+    - main
```
timbastin: The postgresql image does actually not change. that often. I think it is enough to just have it being build on tag

timbastin (approved):

Issue #82, opened 2024-05-13
API Documentation needed 


seb-kw:
Related to: #78 

Requirement (currently marked as unmet): 
> The project MUST provide reference documentation that describes the external interface (both input and output) of the software produced by the project.

I suggest we provide a documentation based on an OpenAPI spec.

Liamhhhhh:
Documents based on the OpenAPI specification can be automatically generated using [Apidog](https://www.apidog.com/?utm_source=&utm_medium=blogger&utm_campaign=test1), and I think the UI looks great. Just a small suggestion from my experience. 
﻿ 
﻿

seb-kw:
Thank you, @Liamhhhhh, for your input — much appreciated! Simplification and automation are (almost) always a good thing 😃 -  we'll take a look at it!

Do you know of any comparable alternatives from your experience? We try to rely exclusively on other open-source projects — after a quick look at Apidog, it looked more like closed source (but please don't hesitate to correct me here).

arne-kroeger:
Hi @seb-kw,

why not using Swaggo (Swagger for Go) to generate the documentation automatically during pipeline and provide the up-2-date version directly for external clients?

seb-kw:
Thanks @arne-kroeger ! 

Looks like a cool option - we'll give it a try here: #94

PR #94, opened 2024-06-06
82 api documentation needed

seb-kw:
Fixes: https://github.com/l3montree-dev/flawfix/issues/82

As suggested in https://github.com/l3montree-dev/flawfix/issues/82#issuecomment-2150285858 we are testing [swaggo](https://github.com/swaggo/swag) for api spec/ doc generation. 

So far:

- Misc route `/api/v1/health`
- Misc route `/api/v1/whoami`
- `/api/v1/vulndb`
- `/api/v1/vulndb/{cveId}`


seb-kw:
@timbastin & @Refoo0 can you please take a look at how you find this way of creating the API documentation and provide feedback?

Good example: https://github.com/l3montree-dev/flawfix/blob/2b65497e3b010f72b16d1769f849e92ecb7ed6ff/internal/core/vulndb/vulndb_controller.go#L34-L47

timbastin:
I think it is fine! We can do it that way

PR #105, opened 2024-06-24
98 replace personal access token random string with a private key we can use that for signing etc

Refoo0:


cmd/flawfind/main.go
```diff
@@ -216,7 +217,13 @@ func init() {
 				slog.Error("could not create request", "err", err)
 				return
 			}
-			req.Header.Set("Authorization", "Bearer "+token)
+
+			err = pat.DoCryptoChallenge(token, req)
```
timbastin: I think we can rename this to "SignRequest". I know, DoCryptoChallenge was my Idea, but since we are now using an official RFC, we can specify this a bit more.

internal/auth/session_middleware.go
```diff
@@ -47,29 +47,15 @@ func cookieAuth(ctx context.Context, oryApiClient *client.APIClient, oryKratosSe
...
 			var userID string
 
 			if oryKratosSessionCookie == nil {
-				// check for authorization header
-				authorizationHeader := c.Request().Header.Get("Authorization")
-				if authorizationHeader == "" {
-					return c.JSON(401, map[string]string{"error": "no session, missing authorization header"})
-				}
-				userID, err = tokenAuth(tokenRepository, authorizationHeader)
+				userID, err = verifier.VerifyCryptoChallenge(c.Request())
```
timbastin: This can be renamed as well to make it simpler to understand what is really going on. `VerifyRequestSignature`

internal/core/pat/pat_service.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+		D: privKeyD,
+	}
+	privKey.PublicKey.X, privKey.PublicKey.Y = privKey.PublicKey.Curve.ScalarBaseMult(privKey.D.Bytes())
+
+	pubKey := &privKey.PublicKey
+	return *pubKey, nil
+}
+
+func PubKeyToFingerprint(pubKey string) (string, error) {
```
timbastin: I think this method should not be exported - as I see, you are currently only using it inside the pat package.

timbastin (approved):

PR #107, opened 2024-06-27
23 daily updated cvss 40 score calculation

Refoo0:


internal/core/flaw/flaw_service.go
```diff
@@ -111,21 +112,59 @@ func (s *service) UserDetectedFlaws(tx core.DB, userID string, flaws []models.Fl
...
+	})
+
+	if err != nil {
+		return fmt.Errorf("could not recalculate raw risk assessment: %v", err)
+	}
+	return nil
+
+}
+
+func (s *service) RecalculateRawRiskAssessment(tx core.DB, userID string, flaws []models.Flaw, justification string, asset models.Asset) error {
```
timbastin, line 154: if tx is nil, a new transaction should be created. it is crucial, that the flaw risk update and the flawevent creation runs inside the same transaction. It is allowed, that this transaction is a parent transaction, since this will force the two sql statements to run inside the same transaction as well.

internal/core/flaw/flaw_service.go
```diff
@@ -111,21 +112,59 @@ func (s *service) UserDetectedFlaws(tx core.DB, userID string, flaws []models.Fl
 	return s.flawEventRepository.SaveBatch(tx, events)
 }
 
-func (s *service) RecalculateRawRiskAssessment(tx core.DB, userID string, flaws []models.Flaw, justification string, asset models.Asset) error {
+func (s *service) RecalculateRawRiskAssessmentSystem() error {
```
timbastin:
```suggestion
func (s *service) RecalculateAllRawRiskAssessments() error {
```

timbastin (approved):

PR #108, opened 2024-07-02
102 scrape exploit pocs from GitHub

Refoo0:


.gitignore
```diff
@@ -10,4 +10,5 @@ csv
 datascience.ipynb
 bom.xml
 .kratos/kratos.yml
-*.private-key.pem
\ No newline at end of file
+*.private-key.pem
+exploits.zip
```
timbastin:
```suggestion
```

Lets do it all in memory - we wont need to create a file.

internal/core/vulndb/exploitdb_service.go
```diff
@@ -53,13 +53,17 @@ func (s *exploitDBService) fetchCSV(ctx context.Context) error {
...
-			continue
-		}
+		/*
+			id, err := strconv.ParseInt(record[0], 10, 64)
+			if err != nil {
+				slog.Warn("could not parse exploit id", "id", id)
+				continue
+			}
+		*/
+		id := record[0]
```
timbastin: I would love to have both ids prefixed - this way, we know where the exploit comes from.

```suggestion
		id := fmt.Sprintf("exploitdb:%s", record[0])
```

internal/core/vulndb/github_exploit_service.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	// Unzip the file
+	zipListing, err := zip.OpenReader("exploits.zip")
+	if err != nil {
+		slog.Info("Error while unzipping the file %s", err)
+		return err
+	}
+
+	defer zipListing.Close()
```
timbastin:
```suggestion
	b, err := io.ReadAll(res.Body)
	if err != nil {
		slog.Info("error while reading the response body", "err", err)
		return err
	}

	defer res.Body.Close()

	// Unzip the file
	zipListing, err := zip.NewReader(bytes.NewReader(b), int64(len(b)))
	if err != nil {
		slog.Info("Error while unzipping the file %s", "err", err)
		return err
	}
```

We need all of the bytes in memory - since we are trying to unzip, which is not possible using a stream of bytes. Thus we can just keep the bytes in memory and unzip them.

This way, we avoid a lot of disk hits for writing and reading the file.

internal/core/vulndb/github_exploit_service.go
```diff
@@ -0,0 +1,166 @@
...
+			// parse the json file
+			err = json.Unmarshal(byteValue, &tmp)
+			if err != nil {
+				slog.Info("Error while parsing the json file %s", err)
+				return err
+			}
+
+			for _, exploit := range tmp {
+
+				id := "github" + strconv.Itoa(exploit.LocalID)
```
timbastin:
```suggestion
				id := "github:" + strconv.Itoa(exploit.LocalID)
```

timbastin: This way we even have a separator for the prefix! Even if we sync another exploitdb which uses strings as ids, we can identify the source.

internal/core/vulndb/github_exploit_service.go
```diff
@@ -0,0 +1,166 @@
...
+		slog.Info("Error while unzipping the file %s", err)
+		return err
+	}
+
+	defer zipListing.Close()
+
+	// Read the json files and save them to the database
+
+	//make exploits map
+	exploits := make(map[string]models.Exploit, 0)
```
timbastin, line 69: That is not an issue right now, but just to mention it: 

This introduces pretty tight coupling between the models.Exploit datatype (our beloved Data-Type, Crafted by hand, only containing hand-picked attributes, which fits perfectly into our great application) and the "github exploit datatype" (whatever that it, I have no idea when people are going to update that datatype, It does contain plenty of attributes we aren't even concerned about, we have no control over it)

From a software architecture point of view it would be better to decouple both datatypes and create functions to convert one into the other.

internal/database/models/exploit_model.go
```diff
@@ -17,19 +17,64 @@ package models
...
-	Updated     *time.Time `json:"updated" gorm:"type:date;"`
+	ID          string     ` gorm:"primaryKey;"`
+	LocalID     int        `json:"id" gorm:"-"`
+	Published   *time.Time `json:"pushed_at" gorm:"type:date;"`
+	Updated     *time.Time `json:"updated_at" gorm:"type:date;"`
 	Author      string     `json:"author" gorm:"type:text;"`
 	Type        string     `json:"type" gorm:"type:text;"`
 	Verified    bool       `json:"verified" gorm:"type:boolean;"`
-	SourceURL   string     `json:"sourceURL" gorm:"type:text;"`
+	SourceURL   string     `json:"html_url" gorm:"type:text;"`
```
timbastin: aha - there is the tight coupling. You changed a great attribute name (sourceURL) to html_url. It would be better to try to decouple both datatypes. You can do that by:

1. Create a new datatype which pretty much looks exactly like the exploit struct, but we love to copy code when we do so for a greater good, like architecture
2. Do not even create a struct and parse the github exploit json into a map.

I prefer 1. 🙂

internal/database/models/exploit_model.go
```diff
@@ -17,19 +17,64 @@ package models
...
+type Owner struct {
+	Login string `json:"login"`
 }
 
 func (m Exploit) TableName() string {
 	return "exploits"
 }
+
+/*
+// Exploit represents a GitHub exploit
```
timbastin: this is a great comment! Please keep that. 

```suggestion
// GitHub exploit json
```

internal/database/models/exploit_model.go
timbastin: https://github.com/l3montree-dev/flawfix/pull/108#discussion_r1662624778

timbastin (approved):

PR #128, opened 2024-07-19
116 updating an organization

Refoo0:


internal/core/org/org_dto.go
```diff
@@ -34,11 +34,11 @@ type createRequest struct {
 }
 
 func (c createRequest) toModel() models.Org {
+
 	return models.Org{
 		Name:                   c.Name,
 		ContactPhoneNumber:     c.ContactPhoneNumber,
 		NumberOfEmployees:      c.NumberOfEmployees,
-		Country:                c.Country,
```
timbastin: I think we need this attribute

timbastin (approved):

PR #155, opened 2024-08-08
129 add dashboard overview page for assets

Refoo0:


Dockerfile
```diff
@@ -22,9 +22,13 @@ RUN go mod download
 RUN CGO_ENABLED=0 go build -o /go/bin/app cmd/devguard/main.go
 RUN CGO_ENABLED=0 go build -o /go/bin/devguard-cli cmd/devguard-cli/main.go
 
+<<<<<<< HEAD
```
timbastin: you commited a merge conflict 😄

cmd/devguard/api/api.go
```diff
@@ -370,6 +377,8 @@ func Start(db core.DB) {
 	assetRouter.GET("/sbom.json/", assetController.SBOMJSON)
 	assetRouter.GET("/sbom.xml/", assetController.SBOMXML)
 
+	assetRouter.GET("/overview/", statisticsController.Overview)
```
timbastin: This looks a bit like you are creating a deep coupling here. The page in the frontend is called overview and now you are creating an api endpoint which is called overview as well. The api endpoint should be independent. I would prefer something like "stats" or alike. Just wanted to point out the independence between frontend pages and api endpoints.

internal/core/asset/asset_service.go
```diff
@@ -239,7 +241,7 @@ func buildBomRefMap(bom normalize.SBOM) map[string]cdx.Component {
 
 func (s *service) UpdateSBOM(asset models.Asset, scanType string, currentVersion string, sbom normalize.SBOM) error {
 	// load the asset components
-	assetComponents, err := s.componentRepository.LoadAssetComponents(nil, asset, scanType, currentVersion)
+	AssetComponents, err := s.componentRepository.LoadAssetComponents(nil, asset, scanType, currentVersion)
```
timbastin: camelCase please...

image

```suggestion
	assetComponents, err := s.componentRepository.LoadAssetComponents(nil, asset, scanType, currentVersion)
```

internal/core/statistics/statistics_controller.go
```diff
@@ -0,0 +1,86 @@
...
+func NewHttpController(assetService statisticsService) *httpController {
+	return &httpController{
+		assetService: assetService,
+	}
+}
+
+func (c *httpController) Overview(ctx core.Context) error {
+	asset := core.GetAsset(ctx)
+
+	t := time.Now().AddDate(0, 0, -30)
```
timbastin: maybe add a comment right here. Looks like "last month"

internal/core/statistics/statistics_controller.go
```diff
@@ -0,0 +1,86 @@
...
+	assetRisksSummary, err := c.assetService.GetAssetFlawsStatistics(asset.ID.String())
+	if err != nil {
+		return fmt.Errorf("Error getting asset risk summary: %v", err)
+	}
+	assetRiskDistribution, err := c.assetService.GetAssetFlawsDistribution(asset.ID.String())
+	if err != nil {
+		return fmt.Errorf("Error getting asset risk distribution: %v", err)
+	}
+
+	assetRisks, err := c.assetService.GetAssetRecentRisksByAssetId(asset.ID)
```
timbastin: All of the different "GetAsset*" calls are data independent. You can do them all in parallel.

---
Some of the new assetService methods are expecting a string, other ones are expecting a uuid. Lets stick to one. I do not care which one 😄

internal/core/statistics/statistics_dto.go
```diff
@@ -0,0 +1,65 @@
...
+	AssetComponents             []models.AssetComponents           `json:"assetComponents"`
+	FlawEvents                  []FlawEventWithFlawNameDTO         `json:"flawEvents"`
+}
+
+type FlawEventWithFlawNameDTO struct {
+	flaw.FlawEventDTO
+	FlawName string `json:"flawName"`
+}
+
+func OverviewToDto(o models.Overview) OverviewDTO {
```
timbastin: I am pretty sure, this does not need to be exported.

```suggestion
func overviewToDto(o models.Overview) OverviewDTO {
```

internal/core/statistics/statistics_dto.go
```diff
@@ -0,0 +1,65 @@
+package statistics
+
+import (
+	"github.com/l3montree-dev/devguard/internal/core/flaw"
+	"github.com/l3montree-dev/devguard/internal/database/models"
+)
+
+type OverviewDTO struct {
```
timbastin: does this really need to be exported?

```suggestion
type overviewDTO struct {
```

internal/core/statistics/statistics_dto.go
```diff
@@ -0,0 +1,65 @@
...
+	AssetRiskDistribution       []models.AssetRiskDistribution     `json:"assetRiskDistribution"`
+	AssetRecentRisks            []models.AssetRecentRisks          `json:"assetRisks"`
+	AssetFlaws                  []models.AssetFlaws                `json:"assetFlaws"`
+	AssetFlawsStateStatistics   models.AssetFlawsStateStatistics   `json:"assetFlawsStateStatistics"`
+	AssetHighestDamagedPackages []models.AssetComponents           `json:"assetHighestDamagedPackages"`
+	AssetComponents             []models.AssetComponents           `json:"assetComponents"`
+	FlawEvents                  []FlawEventWithFlawNameDTO         `json:"flawEvents"`
+}
+
+type FlawEventWithFlawNameDTO struct {
```
timbastin: does this really need to be exported?

```suggestion
type flawEventWithFlawNameDTO struct {
```

internal/core/statistics/statistics_dto.go
```diff
@@ -0,0 +1,65 @@
...
+	flawEvents := o.FlawEvents
+	flawEventDTO := make([]FlawEventWithFlawNameDTO, len(o.FlawEvents))
+	for i, flawEvent := range flawEvents {
+		flawEventDTO[i] = flawEventToDto(flawEvent)
+	}
+
+	return OverviewDTO{
+		TotalDependencies:           o.TotalDependencies,
+		TotalCriticalDependencies:   o.TotalCriticalDependencies,
+		AssetCombinedDependencies:   o.AssetCombinedDependencies,
```
timbastin: Some variables are prefixed with "Asset" like "AssetCombinedDependencies" and some are not like "TotalDependencies". I would prefer having all of them without the "Asset" prefix

internal/core/statistics/statistics_service.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func (s *service) getAssetCriticalDependenciesGroupedByScanType(asset_ID string) ([]models.AssetCriticalDependencies, error) {
+	assets, err := s.statisticsRepository.GetAssetCriticalDependenciesGroupedByScanType(asset_ID)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range assets {
+		assets[i].ScannerID = scanTypeFromScannerID(assets[i].ScannerID)
```
timbastin: Now we will never know if scannerId does contain a scanner id or a scanType. You are reusing an already defined variable to just store some other data there. I am pretty sure it won't take too long until we forget this small hack and the first bug occurs. Let's just keep it as scanner id. If the frontend does not like the long string and wants to display another one instead, it should do so, but this isn't the APIs responsibility. There we can include the conversion right into the view, making it explicit.

internal/core/statistics/statistics_service.go
```diff
@@ -0,0 +1,301 @@
...
+		riskMin := 99.0
+		dayOfRisk := "9999-99-99 00:00:00.000000 +0200 CEST"
+
+		for i := range assetRisk {
+			arbitraryJsonData := make(map[string]interface{})
+			err := json.Unmarshal([]byte(assetRisk[i].ArbitraryJsonData), &arbitraryJsonData)
+			if err != nil {
+				slog.Error("could not parse additional data", "err", err, "flawId", assetRisk[i].FlawID)
+			}
+			risk := arbitraryJsonData["risk"].(float64)
```
timbastin: I think you can just use the RawRiskAssessment?

internal/core/statistics/statistics_service.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	return nil
+
+}
+
+func (s *service) GetAssetRecentRisksByAssetId(assetID uuid.UUID) ([]models.AssetRecentRisks, error) {
+	return s.assetRecentRiskRepository.GetAssetRecentRisksByAssetId(assetID)
+}
+
+func (s *service) GetAssetFlawsDistribution(asset_ID string) ([]models.AssetRiskDistribution, error) {
```
timbastin: camelCase please...

internal/core/statistics/statistics_service.go
```diff
@@ -0,0 +1,301 @@
...
+		assets[i].ScannerID = scanTypeFromScannerID(assets[i].ScannerID)
+	}
+
+	return assets, nil
+
+}
+
+func (s *service) GetAssetFlaws(assetID uuid.UUID) ([]models.AssetFlaws, models.AssetFlawsStateStatistics, []models.AssetComponents, []models.AssetComponents, []models.FlawEventWithFlawName, error) {
+
+	flaws, err := s.statisticsRepository.GetFlawsDetailsByAssetId(assetID)
```
timbastin:
```suggestion
	flaws, err := s.statisticsRepository.GetFlawDetailsByAssetId(assetID)
```

internal/database/models/statistic_model.go
```diff
@@ -0,0 +1,86 @@
...
+
+type AssetRiskDistribution struct {
+	ScannerID string `json:"scannerId"`
+	RiskRange string `json:"riskRange"`
+	Count     int64  `json:"count"`
+}
+
+type AssetRecentRisks struct {
+	AssetID      uuid.UUID `json:"assetId" gorm:"primaryKey"`
+	ID           int       `json:"id" gorm:"primaryKey"`
```
timbastin: is the id really necessary. Maybe we can just have the DayOfRisk as primary key

internal/database/models/statistic_model.go
```diff
@@ -0,0 +1,86 @@
...
+	DayOfRisk    string    `json:"dayOfRisk" `
+	DayOfScan    string    `json:"dayOfScan"`
```
timbastin: What is the difference between dayOfRisk and dayOfScan? Maybe add a comment right here

internal/database/models/statistic_model.go
```diff
@@ -0,0 +1,86 @@
...
+	AssetSumRisk float64   `json:"assetSumRisk"`
```
timbastin: Now you have it double prefixed. AssetRecentRisks.AssetSumRisk. I think everyone already knows its about assets. how about: AssetRecentRisks.SumRisk

internal/database/repositories/component_repository.go
```diff
@@ -104,3 +104,33 @@ func (c *componentRepository) HandleStateDiff(tx database.DB, assetID uuid.UUID,
...
+		return nil, err
+	}
+
+	return results, nil
+}
+
+func (c *componentRepository) GetPackages(asset_ID string) ([]models.AssetComponents, error) {
+	var results []models.AssetComponents
+	err := c.db.Model(&models.Component{}).
+		Select("SUBSTRING(purl_or_cpe FROM ':(.*?)/') AS pkg , COUNT(*) as count").
```
timbastin: why are you doing that? Maybe add a comment

internal/database/repositories/statistics_repository.go
```diff
@@ -0,0 +1,173 @@
...
+			flaw_id;
+	`, time, assetID).Scan(&flawRisk).Error; err != nil {
+
+		return nil, err
+	}
+
+	return flawRisk, nil
+}
+
+func (r *statisticsRepository) GetAssetCriticalDependenciesGroupedByScanType(asset_ID string) ([]models.AssetCriticalDependencies, error) {
```
timbastin: I think the name is misleading. You are returning all just a scanner_id and a count. Basically the amount of dependencies per scan type. You are not returning the critical dependencies of the asset - thats what the name implies

PR #245, opened 2025-01-28
started to refactor: rename Asset to AssetNew and update related refe…

refoo0:
…rence


timbastin:
```sql
DELETE FROM asset_versions;

INSERT INTO asset_versions (name, asset_id, slug, type, default_branch) 
SELECT 'main' as name, id as asset_id, 'main' as slug, 'branch' as type, true as default_branch from assets 
ON CONFLICT (name, asset_id) DO NOTHING


ALTER TABLE flaws ADD asset_version_name text;
UPDATE flaws SET asset_version_name = 'main';
ALTER TABLE flaws ALTER COLUMN asset_version_name SET NOT NULL;
```

PR #261, opened 2025-02-24
251 button to upload an SBOM manually

Hubtrick-Git:
Added an api interface where it takes an sbom file and runs a scan over it. The function "ManualSbomScan" is called using the api.go file and lies in the scan_controller.go file where it re-uses most of the code from the "Scan"  function.
The function then returns the scan results as well as potential errors while scanning.

### DIFF
--- a/cmd/devguard/api/api.go
+++ b/cmd/devguard/api/api.go
@@ -511,7 +511,7 @@ func Start(db core.DB) {
 	assetRouter.GET("/refs/", assetVersionController.GetAssetVersionsByAssetID)
 
 	//Api to scan manually using an uploaded SBOM provided by the user
-	assetRouter.POST("/sbom-manual-scan/", assetController.ManualSbomScan)
+	assetRouter.POST("/sbom-manual-scan/", scanController.ManualSbomScan)
 
 	//TODO: add the projectScopedRBAC middleware to the following routes
 	assetVersionRouter := assetRouter.Group("/refs/:assetVersionSlug", assetVersionMiddleware(assetVersionRepository))
--- a/cmd/devguard/api/api.go
+++ b/cmd/devguard/api/api.go
@@ -339,10 +339,6 @@ func health(c echo.Context) error {
 	return c.String(200, "ok")
 }
 
-func manualSbomScan() {
-
-}
-
 func Start(db core.DB) {
 
 	ory := auth.GetOryApiClient(os.Getenv("ORY_KRATOS_PUBLIC"))
--- a/cmd/devguard/api/api.go
+++ b/cmd/devguard/api/api.go
@@ -507,7 +507,7 @@ func Start(db core.DB) {
 	assetRouter.GET("/refs/", assetVersionController.GetAssetVersionsByAssetID)
 
 	//Api to scan manually using an uploaded SBOM provided by the user
-	assetRouter.POST("/sbom-manual-scan/", scanController.ManualSbomScan)
+	assetRouter.POST("/sbom-file/", scanController.ScanSbomFile)
 
 	//TODO: add the projectScopedRBAC middleware to the following routes
 	assetVersionRouter := assetRouter.Group("/refs/:assetVersionSlug", assetVersionMiddleware(assetVersionRepository))
--- a/cmd/devguard/api/api.go
+++ b/cmd/devguard/api/api.go
@@ -339,6 +339,10 @@ func health(c echo.Context) error {
 	return c.String(200, "ok")
 }
 
+func manualSbomScan() {
+
+}
+
 func Start(db core.DB) {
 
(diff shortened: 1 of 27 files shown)


internal/core/vulndb/scan/scan_controller.go
```diff
@@ -193,3 +194,122 @@ func (s *httpController) Scan(c core.Context) error {
...
+			return c.JSON(500, map[string]string{"error": "could not recalculate risk history"})
+		}
+
+		// save the asset
+		if err := s.assetVersionRepository.Save(nil, &assetVersion); err != nil {
+			slog.Error("could not save asset", "err", err)
+		}
+	}
+
+	file.Close() //Close file to prevent memory leak
```
timbastin: I would really prefer a `defer` function for that - right after extracting it from the multipart request.

Line 211: `defer file.Close()`

timbastin (changes requested):
I think there is some code duplication in the scan_controller.go right now. Maybe a non exported method on the scanController can be created which does the duplicated parts. What do you think?

internal/core/vulndb/scan/scan_controller.go
```diff
@@ -90,14 +91,8 @@ type ScanResponse struct {
...
 
-func (s *httpController) Scan(c core.Context) error {
-	bom := new(cdx.BOM)
-	decoder := cdx.NewBOMDecoder(c.Request().Body, cdx.BOMFileFormatJSON)
-	if err := decoder.Decode(bom); err != nil {
-		return err
-	}
-	normalizedBom := normalize.FromCdxBom(bom, true)
-
+func ScanHelpFunction(c core.Context, bom normalize.SBOM, s *httpController) error {
```
timbastin: We should keep the exported functions to a minimum. Since this is a helper function which only gets used in the same package, it it fine to not export it. By defining it as a method on the httpController you do not need to pass it as parameter. This is just a bit more elegant.

```suggestion
func (s *httpController) scanSBOM(c core.Context, bom normalize.SBOM) error {
```

By renaming it, it is simply a bit more expressive. `HelperFunction` could be everything and nothing.

cmd/devguard/api/api.go
```diff
@@ -497,12 +499,16 @@ func Start(db core.DB) {
...
 
+	//Api functions for interacting with an asset inside a project  ->  .../projects/<project-name>/assets/<asset-name>/...
 	assetRouter := projectRouter.Group("/assets/:assetSlug", projectScopedRBAC(accesscontrol.ObjectAsset, accesscontrol.ActionRead), assetMiddleware(assetRepository))
 	assetRouter.GET("/", assetController.Read)
 	assetRouter.DELETE("/", assetController.Delete, projectScopedRBAC(accesscontrol.ObjectAsset, accesscontrol.ActionDelete))
 
 	assetRouter.GET("/refs/", assetVersionController.GetAssetVersionsByAssetID)
 
+	//Api to scan manually using an uploaded SBOM provided by the user
+	assetRouter.POST("/sbom-manual-scan/", scanController.ManualSbomScan)
```
```suggestion
	assetRouter.POST("/sbom-file", scanController.ScanSbomFile)
```

Lets rename that. Since the backend does not know, that the sbom was manually uploaded, i think removing this suffix makes sense.

internal/core/vulndb/scan/scan_controller.go
```diff
@@ -192,4 +187,48 @@ func (s *httpController) Scan(c core.Context) error {
 		AmountClosed: amountClose,
```
timbastin: Your ScanHelperFunction is already returning an HTTP-Response by Calling c.JSON. Lets keep HTTP-Request handling in the "Real" Controller methods, which get called by the http server handler. I would prefer you to return the ScanResponse right here, and then calling c.JSON in the `Scan` and `ScanSbomFile` method. What do you think? Currently you are calling c.JSON twice, which wont work I am afraid

internal/core/vulndb/scan/scan_controller.go
```diff
@@ -192,4 +187,48 @@ func (s *httpController) Scan(c core.Context) error {
...
+
+	err := ScanHelpFunction(c, normalize.FromCdxBom(bom, true), s)
+
+	return c.JSON(200, err)
+
+}
+
+func (s *httpController) ManualSbomScan(c core.Context) error {
+
+	var maxSize int = 16 * 1024 * 1024 //Max Upload Size 16mb
```
timbastin:
```suggestion
	var maxSize int64 = 16 * 1024 * 1024 //Max Upload Size 16mb
```

Thsi way you avoid the casting in line 212

timbastin (changes requested):
Great Job! I think there are just a few further adjustments to make - especially renaming functions.

timbastin (approved):

PR #278, opened 2025-02-27
267 cannot create new asset version after deleting one with the same name

Hubtrick-Git:
Added the requested fixes as well as a testing function ( which is half baked because of the missing context)


internal/database/repositories/asset_version_repository.go
```diff
@@ -84,7 +85,12 @@ func (a *assetVersionRepository) FindOrCreate(assetVersionName string, assetID u
 		}
 
 		if err = a.db.Create(&models.AssetVersion{Name: assetVersionName, AssetID: assetID, Slug: assetVersionName, Type: assetVersionType, DefaultBranch: defaultBranch}).Error; err != nil {
+			if strings.Contains(err.Error(), "duplicate key value violates") { //Check if the error is due to duplicate keys
```
timbastin: I dislike your comments. Try to explain the "why" and not the "what". The "what" I can already see in that line: `strings.Contains(err.Error(), "duplicate key value violates")`. Code is super expressive: I know exactly what you are doing there.

But why are you doing it? Why checking for duplicate key values? **Because this means we are trying to create an asset which does already exist**. That would be a better comment 🙂

internal/database/repositories/asset_version_repository.go
```diff
@@ -84,7 +85,12 @@ func (a *assetVersionRepository) FindOrCreate(assetVersionName string, assetID u
...
+				a.db.Unscoped().Model(&app).Where("name", assetVersionName).Update("deleted_at", nil) //Update deleted at to NULL
```
timbastin:
```suggestion
				a.db.Unscoped().Model(&app).Where("name", assetVersionName).Update("deleted_at", nil) // Update deleted at to NULL is gorms way of returning a soft deleted record to the active records. There is no special function for that. Note the unscoped method usage
```

timbastin (changes requested):

timbastin:
I think the testing of this function is really hard since it is on such a deep layer just interacting with the database. Testing repositories is awful since there is such a deep coupling with the `gorm` database. You would need to mock the whole `db *gorm.DB` struct to make that possible.

timbastin:
Your tests still fail. Just remove them altogether 🙂 I do not really get the changes in the docker-compose.yaml. I think those need to be reverted since it is not possible to have image and build defined on the same serve i am afraid.

Hubtrick-Git:
Added requested changes to the code. Ready to merge now

PR #279, opened 2025-02-28
No more CVE duplicates in the risk handling scan overview

Hubtrick-Git:
Added a check when building the Flaw List if the flaw is already present using a Slice.
For every flaw : slice.contains get called so for m = len(slice) and n = number of flaws Time Complexity is O(n * m) <  O(n^2)
Could be reduced to O(n) time complexity with using a hashmap maybe


timbastin:
I am afraid the solution is much more complicated. I added some comments right here: https://github.com/l3montree-dev/devguard/issues/197#issuecomment-2692666967

I am closing this merge request right now, since I think we need to have some more discussion about the possible solution in the first place.

PR #288, opened 2025-03-04
282 whole system crash when sending empty gitlab string

Hubtrick-Git:
Fixed the Issue by checking for an empty token and added a test for it


internal/core/integrations/gitlab_integration.go
```diff
@@ -512,8 +513,14 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
...
 
+	chosenURL := os.Getenv("INSTANCE_DOMAIN")
+	if chosenURL == "" {
+		slog.Error("No URL specified in .env file")
+		return nil
+	}
+
 	for _, hook := range hooks {
-		if hook.URL == "https://api.main.devguard.org/api/v1/webhook/" {
+		if hook.URL == chosenURL {
```
timbastin: I prefer passing the domain without any path. Like: "https://api.main.devguard.org"

```suggestion
		if strings.HasPrefix(hook.URL, chosenURL) {
```

internal/core/integrations/gitlab_integration.go
```diff
@@ -512,8 +513,14 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
 		return fmt.Errorf("could not list project hooks: %w", err)
 	}
 
+	chosenURL := os.Getenv("INSTANCE_DOMAIN")
```
timbastin:
```suggestion
	instanceDomain := os.Getenv("INSTANCE_DOMAIN")
```

internal/core/integrations/gitlab_integration.go
```diff
@@ -556,6 +558,36 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
 
 }
 
+func CreateProjectHookOptions(token uuid.UUID, hooks []*gitlab.ProjectHook) (*gitlab.AddProjectHookOptions, error) {
```
timbastin: You do not need to export this function

```suggestion
func createProjectHookOptions(token uuid.UUID, hooks []*gitlab.ProjectHook) (*gitlab.AddProjectHookOptions, error) {
```

internal/core/integrations/gitlab_integration_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			slog.Error("error when trying to generate token")
+			return
+		}
+		results, err := CreateProjectHookOptions(token, hooks)
+		if err != nil {
+			slog.Error(err.Error())
+			return
+		}
+		assert.Equal(t, "https://api.stage.devguard.org", *results.URL)
```
timbastin: We need the whole webhook url (including the path) to be added. This applies to all tests
```suggestion
		assert.Equal(t, "https://api.stage.devguard.org/api/v1/webhook/", *results.URL)
```

internal/core/integrations/gitlab_integration.go
```diff
@@ -556,6 +558,36 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
...
+	projectOptions := &gitlab.AddProjectHookOptions{} //Intialize empty struct to return on error
+
+	chosenURL := os.Getenv("INSTANCE_DOMAIN") //Get the URL from the .env file
+
+	for _, hook := range hooks { //Check if the Hook already exists
```
timbastin: do this check in the calling function. Actually we do not want todo anything if the project hook already exists. This way, we would create an empty hook, since this method does not return any error

internal/core/integrations/gitlab_integration.go
```diff
@@ -556,6 +558,36 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
...
+		}
+	}
+
+	//Assign values to the empty struct
+	projectOptions.IssuesEvents = gitlab.Ptr(true)
+	projectOptions.ConfidentialIssuesEvents = gitlab.Ptr(true)
+	projectOptions.NoteEvents = gitlab.Ptr(true)
+	projectOptions.ConfidentialNoteEvents = gitlab.Ptr(true)
+	projectOptions.EnableSSLVerification = gitlab.Ptr(true)
+	if chosenURL == "" { //If no URL is provided in the enviroment variables default to main URL
```
timbastin: I like this logic - but it will never run, since the calling function returns an error if instanceDomain is empty

internal/core/project/project_controller.go
```diff
@@ -347,7 +347,8 @@ func (p *Controller) Delete(c core.Context) error {
 }
 
 func (p *Controller) Read(c core.Context) error {
-	// just get the project from the context
+	// just get the project from the contextx
+	fmt.Printf("-----------This is the URL: %s----------------\n\n", c.Get("url"))
```
timbastin: this can be deleted 🙂

timbastin (changes requested):

Hubtrick-Git:
Fixed error in Code, requested by the code review. ready to merge

cmd/devguard/main.go
```diff
@@ -45,7 +48,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-
+	fmt.Printf("This is the env: %s", os.Getenv("INSTANCE_DOMAIN"))
```
timbastin: this can be deleted :)

internal/core/integrations/gitlab_integration.go
```diff
@@ -556,6 +545,37 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
...
 }
 
+func createProjectHookOptions(token uuid.UUID, hooks []*gitlab.ProjectHook) (*gitlab.AddProjectHookOptions, error) {
+	projectOptions := &gitlab.AddProjectHookOptions{} //Intialize empty struct to return on error
+
+	instanceDomain := os.Getenv("INSTANCE_DOMAIN") //Get the URL from the .env file
+
+	for _, hook := range hooks {
+		if strings.HasPrefix(hook.URL, instanceDomain) {
+			slog.Error("hook already exists")
```
timbastin: I think this log statement is not really valuable. We do not know which organization, asset nor project its about. I think it is safe to just delete this.

internal/core/integrations/gitlab_integration.go
```diff
@@ -556,6 +545,37 @@ func (g *gitlabIntegration) addProjectHook(ctx core.Context) error {
...
+	}
+
+	//Assign values to the empty struct
+	projectOptions.IssuesEvents = gitlab.Ptr(true)
+	projectOptions.ConfidentialIssuesEvents = gitlab.Ptr(true)
+	projectOptions.NoteEvents = gitlab.Ptr(true)
+	projectOptions.ConfidentialNoteEvents = gitlab.Ptr(true)
+	projectOptions.EnableSSLVerification = gitlab.Ptr(true)
+	if instanceDomain == "" { //If no URL is provided in the enviroment variables default to main URL
+		slog.Error("no URL specified in .env file defaulting to main")
```
timbastin:
```suggestion
		slog.Debug("no URL specified in .env file defaulting to main")
```

This is actually not an error. It might even be intended.

internal/core/integrations/gitlab_integration_test.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		results, err := createProjectHookOptions(token, hooks)
+		if err != nil {
+			slog.Error(err.Error())
+			return
+		}
+		assert.Equal(t, "https://api.main.devguard.org/api/v1/webhook/", *results.URL)
+
+	})
+}
```
timbastin, line 107: Awesome!! All of them look great! Lets add another simple test case. I think a lot of users might configure the environment variable with a trailing slash. Does it work then?

```
os.Setenv("INSTANCE_DOMAIN", "https://api.stage.devguard.org/")
```

Should result in `https://api.stage.devguard.org/api/v1/webhook/`

internal/core/integrations/gitlab_integration_test.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		results, err := createProjectHookOptions(token, hooks)
+		if err != nil {
+			slog.Error(err.Error())
+			return
+		}
+		assert.Equal(t, "https://api.main.devguard.org/api/v1/webhook/", *results.URL)
+
+	})
+	t.Run("Returned ProjectHookOption Struct should have the URL set to stage devguard", func(t *testing.T) {
```
timbastin: > Returned ProjectHookOption Struct should have the URL set to stage devguard

Yep. I see that. By why should it? Try to explain the why and not the "what". Same for the other tests.

> Returned ProjectHookOption Struct should have the URL set to stage devguard **since it is set as INSTANCE_DOMAIN environment variable

timbastin (changes requested):
Looks great! Just some smaller changes. Have a look at your pipeline. In this case it wasnt your fault but we need to make sure it does pass before we can merge it.

internal/core/integrations/gitlab_integration_test.go
Hubtrick-Git: Good point, it doesn't work like that
Gonna fix it

Hubtrick-Git:
Changed the Code accordingly and fixed the bug for trailing slashes

timbastin (approved):

PR #290, opened 2025-03-05
fix create asset version bug

refoo0:


internal/database/repositories/asset_version_repository_test.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/l3montree-dev/devguard/internal/database/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAssetVersionFactory(t *testing.T) {
+	repo := &assetVersionRepository{}
```
timbastin: I would really love to have a t.Run statement right here. The only purpose is, that we can pass a string there like:

```
t.Run("should slugify the asset name and store it in the slug property", ...)

timbastin (approved):

PR #306, opened 2025-03-11
289 automatically create ticket for high risk or or high cvss vulnerabilities

Hubtrick-Git:
Integrated function into scan controller to check on each found vulnerability if it exceeds the threshold set by the organization. The same was done for the automatic Updates of the CVE´s.
Heavy performance improvements can be done in the automatic Update Logic


Hubtrick-Git:
So Chef, ist erledigt 🚀

timbastin (approved):

PR #315, opened 2025-03-13
310 server stuck in loop when updating an organization name to an empty name

Hubtrick-Git:
Added checks to the update and create functions of 
-the organization controller
-the project controller
-the asset controller
(asset version repository only on create)


Hubtrick-Git:
Additionally I changed some variable names to better describe them

Hubtrick-Git:
Bug is fixed.
A lot more test coverage can be done in org controller but its probably better to put that on ice for now and do the testing in a separate branch later.

timbastin (approved):

PR #326, opened 2025-03-17
update git command to include 'v' prefix for latest tag

refoo0:


timbastin:
I do not really see how this resolves the issue. The code seems a bit more elegant, but the same in what it is doing. Am I missing something?

timbastin (approved):

PR #332, opened 2025-03-18
293 allow users to specify access scope when generating personal access tokens

refoo0:


cmd/devguard/api/api.go
```diff
@@ -458,7 +475,7 @@ func BuildRouter(db core.DB) *echo.Echo {
 	orgRouter.GET("/", orgController.List)
 
 	//Api functions for interacting with an organization  ->  .../organizations/<organization-name>/...
-	organizationRouter := orgRouter.Group("/:organization", multiOrganizationMiddleware(casbinRBACProvider, orgRepository))
+	organizationRouter := orgRouter.Group("/:organization", multiOrganizationMiddleware(casbinRBACProvider, orgRepository), neededScope([]string{"manageAsset", "scanAsset"}))
```
```suggestion
	organizationRouter := orgRouter.Group("/:organization", multiOrganizationMiddleware(casbinRBACProvider, orgRepository), neededScope([]string{"manageAsset"}))
```

cmd/devguard/api/api.go
```diff
@@ -237,6 +238,22 @@ func projectAccessControl(projectRepository projectRepository, obj accesscontrol
 	}
 }
 
+func neededScope(neededScopes []string) core.MiddlewareFunc {
```
timbastin, line 241: Would be awesome to have a test for this function

cmd/devguard/api/api.go
```diff
@@ -438,9 +455,9 @@ func BuildRouter(db core.DB) *echo.Echo {
...
-	sessionRouter.POST("/scan/", scanController.ScanDependencyVulnFromProject, assetNameMiddleware(), multiOrganizationMiddleware(casbinRBACProvider, orgRepository), projectScopedRBAC(accesscontrol.ObjectAsset, accesscontrol.ActionUpdate), assetMiddleware(assetRepository))
+	sessionRouter.POST("/scan/", scanController.ScanDependencyVulnFromProject, neededScope([]string{"scan"}), assetNameMiddleware(), multiOrganizationMiddleware(casbinRBACProvider, orgRepository), projectScopedRBAC(accesscontrol.ObjectAsset, accesscontrol.ActionUpdate), assetMiddleware(assetRepository))
 
-	sessionRouter.POST("/sarif-scan/", scanController.FirstPartyVulnScan, assetNameMiddleware(), multiOrganizationMiddleware(casbinRBACProvider, orgRepository), projectScopedRBAC(accesscontrol.ObjectAsset, accesscontrol.ActionUpdate), assetMiddleware(assetRepository))
+	sessionRouter.POST("/sarif-scan/", scanController.FirstPartyVulnScan, neededScope([]string{"scan"}), assetNameMiddleware(), multiOrganizationMiddleware(casbinRBACProvider, orgRepository), projectScopedRBAC(accesscontrol.ObjectAsset, accesscontrol.ActionUpdate), assetMiddleware(assetRepository))
 
 	patRouter := sessionRouter.Group("/pats")
```
timbastin: We have a privilege escalation right here. A token with scope "scan" can create new tokens with "manage" access. We need to apply the neededScope middleware here as well.

internal/auth/session_middleware.go
```diff
@@ -58,10 +59,11 @@ func SessionMiddleware(oryApiClient *client.APIClient, verifier verifier) echo.M
 			oryKratosSessionCookie := getCookie("ory_kratos_session", ctx.Cookies())
 
 			var userID string
+			var scopes string
 			var err error
 
 			if oryKratosSessionCookie == nil {
-				userID, err = verifier.VerifyRequestSignature(ctx.Request())
+				userID, scopes, err = verifier.VerifyRequestSignature(ctx.Request())
```
timbastin, line 66: It would be great to have a test for this function here as well. That the correct scopes get set.

internal/core/pat/pat_controller.go
```diff
@@ -63,6 +63,7 @@ func (p *PatController) Create(c core.Context) error {
 		"userId":      patStruct.UserID.String(),
 		"pubKey":      patStruct.PubKey,
 		"fingerprint": patStruct.Fingerprint,
+		"scopes":      patStruct.Scopes,
```
timbastin, line 66: I think we need to validate the scopes which were sent. Lets just create a allowed-scopes list and we check if all the scopes are inside that list

refoo0:
UPDATE pat SET scopes = 'scan' WHERE true;

internal/core/pat/pat_dto.go
```diff
@@ -38,9 +43,18 @@ func (p CreateRequest) ToModel(userID string) models.PAT {
 		slog.Error("could not convert public key to fingerprint", "err", err)
 		return models.PAT{}
 	}
+
+	//check if the scopes are valid
+	ok := utils.ContainsAll(strings.Fields(p.Scopes), AllowedScopes)
```
timbastin: I think this wont work. Doesnt it say, the p.Scopes has to contain ALL allowed scopes. Does it work, if i want to create a token with just "scan" as scope?

timbastin (changes requested):

timbastin (approved):

PR #359, opened 2025-03-24
356 only automatically create tickets if there is no other asset version with an already created ticket for this vulnerability

Hubtrick-Git:
Added logic to create an issue for a given vulnerability only if it was found on the default branch, sourced it out into its own function and wrote a adequate tests for it.

### DIFF
--- a/internal/core/vulndb/scan/scan_controller.go
+++ b/internal/core/vulndb/scan/scan_controller.go
@@ -133,10 +133,12 @@ func DependencyVulnScan(c core.Context, bom normalize.SBOM, s *httpController) (
 		return scanResults, err
 	}
 
-	//Only create Tickets if the assetVersion is the default assetVersion
-	err = shouldCreateIssue(s, asset, assetVersion, newState)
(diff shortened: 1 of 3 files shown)


internal/core/vulndb/scan/scan_controller.go
```diff
@@ -158,6 +159,17 @@ func DependencyVulnScan(c core.Context, bom normalize.SBOM, s *httpController) (
 	return scanResults, nil
 }
 
+func ShouldCreateIssue(s *httpController, asset models.Asset, assetVersion models.AssetVersion, newState []models.DependencyVuln) error {
```
timbastin: You do not need to export this function

timbastin (changes requested):

internal/core/vulndb/scan/scan_controller.go
timbastin: Maybe returning true, false would be simpler

timbastin (approved):

PR #367, opened 2025-03-25
360 devguard scanner link is incorrect

Hubtrick-Git:
Now the assetVersion also gets trasnmitted over the various functions to be used in the embedded link when creating a ticket.


timbastin (changes requested):
You fixed the link in the tickets! Thats great. Could you fix it in the scanner output as well? https://github.com/l3montree-dev/devguard/blob/main/cmd/devguard-scanner/commands/sca.go#L192

PR #372, opened 2025-03-27
Install gitleaks and semgrep in scanner dockerfile

refoo0:


timbastin:
@refoo0 Lets stop any "Multi-Arch"-Effort right here. Did not work well since kaniko doesnt support it. We can get back to it sooner or later!

timbastin:
Can you add this here as well? https://github.com/bridgecrewio/checkov ❤️

install_devguard_scanner_deps.sh
timbastin, line 1: I think we do not need this anymore

timbastin (changes requested):

timbastin (approved):

PR #427, opened 2025-04-01
364 add dependency graph to GitHub and gitlab tickets using mermaidjs

Hubtrick-Git:
Issues now show a mermaid flow chart with the individual components in their description . Could need some artistic and formating improvements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ clean::
 	docker compose down -v && docker compose up -d
 
 mocks::
-	mockery --config=.mockery.yaml 
+	mockery --config=.mockery.yaml
 
 lint::
 	golangci-lint run ./...
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ clean::
 	docker compose down -v && docker compose up -d
 
 mocks::
-	mockery --config=mockery.yaml --all
+	mockery --config=.mockery.yaml 
 
 lint::
 	golangci-lint run ./...
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ clean::
 	docker compose down -v && docker compose up -d
 
 mocks::
-	mockery --config=.mockery.yaml
+	mockery --config=mockery.yaml --all
 
 lint::
 	golangci-lint run ./...
--- a/internal/core/integrations/github_integration.go
+++ b/internal/core/integrations/github_integration.go
@@ -745,7 +745,7 @@ func (g *githubIntegration) CreateIssue(ctx context.Context, asset models.Asset,
 }
 
 func (g *githubIntegration) renderPathToComponent(assetID uuid.UUID, assetVersionName string, scannerID string, pURL string) (string, error) {
-	var mermaidFlowChart string
+	mermaidFlowChart := "mermaid \n %%{init: { 'theme':'dark' } }%%\n flowchart LR\n"
 	components, err := g.componentRepository.LoadPathToComponent(nil, assetVersionName, assetID, pURL, scannerID)
 	if err != nil {
 		return mermaidFlowChart, err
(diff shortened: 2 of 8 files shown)


internal/core/integrations/github_integration.go
```diff
@@ -734,3 +743,40 @@ func (g *githubIntegration) CreateIssue(ctx context.Context, asset models.Asset,
...
+		} else {
+			break
+		}
+	}
+
+	//now we build the string using this list, every new node need prefix and suffix to work with mermaid. [] are used to prohibit mermaid from interpreting some symbols from the package names as mermaid syntax
+	mermaidFlowChart += componentList[0]
+
+	for i, componentName := range componentList[1:] {
+		mermaidFlowChart = mermaidFlowChart + " --> \n" + "node" + strconv.Itoa(i) + "[" + componentName + "]"
```
timbastin:
```suggestion
		mermaidFlowChart = mermaidFlowChart + " --> " + "node" + strconv.Itoa(i) + "[" + componentName + "]"
```

internal/core/integrations/github_integration.go
```diff
@@ -663,9 +668,13 @@ func (g *githubIntegration) CreateIssue(ctx context.Context, asset models.Asset,
 
 	assetSlug := asset.Slug
 	labels := getLabels(&dependencyVuln, "open")
+	componentTree, err := g.renderPathToComponent(asset.ID, assetVersionName, "SBOM-File-Upload", exp.AffectedComponentName)
```
timbastin: I believe you can get the scanner id from the vulnerability

```suggestion
	componentTree, err := g.renderPathToComponent(asset.ID, assetVersionName, dependencyVuln.ScannerID, exp.AffectedComponentName)
```

internal/core/integrations/gitlab_integration.go
```diff
@@ -1066,3 +1074,40 @@ func (g *gitlabIntegration) CreateIssue(ctx context.Context, asset models.Asset,
 
 	return g.dependencyVulnRepository.ApplyAndSave(nil, &dependencyVuln, &vulnEvent)
 }
+
+// this function returns a string containing a mermaids js flow chart to the given pURL
+func (g *gitlabIntegration) renderPathToComponent(assetID uuid.UUID, assetVersionName string, scannerID string, pURL string) (string, error) {
```
timbastin: I think you can reduce the duplication of this logic by creating its own function, which gets the component repository as parameter. And please write a test for it 🙂

timbastin (changes requested):

Hubtrick-Git:
Code is ready .The only thing left to do is write tests

internal/core/integrations/__debug_bin2147138446
timbastin, line 1: this can be deleted 🙂

internal/core/integrations/thirdparty_integration_test.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		componentRepository := mocks.NewCoreComponentRepository(t)
+		componentRepository.On("LoadPathToComponent", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(components, nil)
+
+		assetID := uuid.New()
+		assetVersionName := "TestName"
+		scannerID := "SBOM-File-Upload"
+		pURL := "pkg:npm:test"
+
+		_, err := integrations.RenderPathToComponent(componentRepository, assetID, assetVersionName, scannerID, pURL)
```
timbastin: It would be great if you would do an assertion on the output of the `RenderPathToComponent`. Just check, that the string looks like you are expecting. I am thinking of a simple  `assert.Equal(t, ```mermaidjs....```, actual)`

timbastin (changes requested):

Hubtrick-Git:
Disclaimer I know that the FormatNode function should not be public but i can't use it in the integrations_text package otherwise.

Makefile
```diff
@@ -7,7 +7,7 @@ clean::
 	docker compose down -v && docker compose up -d
 
 mocks::
-	mockery --config=.mockery.yaml
+	mockery --config=mockery.yaml --all
```
timbastin: What are those changes right here? I think you removed the dot accidentally and there you had to add the `--all` flag

internal/core/integrations/thirdparty_integration_test.go
```diff
@@ -0,0 +1,97 @@
...
+		scannerID := "SBOM-File-Upload"
+		pURL := "pkg:npm:test"
+
+		result, err := integrations.RenderPathToComponent(componentRepository, assetID, assetVersionName, scannerID, pURL)
+		if err != nil {
+			t.Fail()
+		}
+
+		//String for the empty graph + 1 node being root with a linebreak
+		assert.Equal(t, "```mermaid \n %%{init: { 'theme':'dark' } }%%\n flowchart LR\nroot\n```\n", result)
```
timbastin: this looks like a really small flowchart. Just the root node?

timbastin:
> Disclaimer I know that the FormatNode function should not be public but i can't use it in the integrations_text package otherwise.

Maybe it is just simpler to change the mermaidjs diagram from `LR` to `TB`?

Makefile
Hubtrick-Git: Yeah sorry that got mixed up when I tried to set everything up on linux, you can ignore that / dont merge it if thats possible or I can revert it.

timbastin (approved):

PR #543, opened 2025-04-22
522 add config file endpoints to asset project and organization

Hubtrick-Git:
added the requested function and the necessary endpoints. 
also made some cosmetic changes : )

### DIFF
--- a/internal/core/asset/asset_controller.go
+++ b/internal/core/asset/asset_controller.go
@@ -222,17 +222,18 @@ func (a *httpController) Update(ctx core.Context) error {
 	return ctx.JSON(200, asset)
 }
 
-func (a *httpController) GetConfigFile(ctx core.Context, configID string) string {
+func (a *httpController) GetConfigFile(ctx core.Context) error {
 	organization := core.GetOrganization(ctx)
 	project := core.GetProject(ctx)
 	asset := core.GetAsset(ctx)
+	configID := ctx.Param("config-file")
 
 	configContent := asset.ConfigFiles[configID]
 	if configContent.(string) == "" {
 		if project.ConfigFiles[configID].(string) == "" {
-			return organization.ConfigFiles[configID].(string)
+			return ctx.JSON(200, organization.ConfigFiles[configID].(string))
 		}
-		return project.ConfigFiles[configID].(string)
+		return ctx.JSON(200, project.ConfigFiles[configID].(string))
 	}
-	return configContent.(string)
+	return ctx.JSON(200, configContent.(string))
 }
--- a/internal/core/asset/asset_controller.go
+++ b/internal/core/asset/asset_controller.go
@@ -229,7 +229,7 @@ func (a *httpController) GetConfigFile(ctx core.Context) error {
 	configID := ctx.Param("config-file")
 
 	configContent := asset.ConfigFiles[configID]
-	if configContent.(string) == "" {
+	if configContent.(string) == "" { //if we have no config files in this asset we want to look in the corresponding project and then in the organization
 		if project.ConfigFiles[configID].(string) == "" {
 			return ctx.JSON(200, organization.ConfigFiles[configID].(string))
 		}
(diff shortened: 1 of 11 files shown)


internal/core/asset/asset_dto.go
```diff
@@ -22,6 +23,8 @@ type createRequest struct {
 	ConfidentialityRequirement string `json:"confidentialityRequirement" validate:"required"`
 	IntegrityRequirement       string `json:"integrityRequirement" validate:"required"`
 	AvailabilityRequirement    string `json:"availabilityRequirement" validate:"required"`
+
+	ConfigFiles database.JSONB `json:"configFiles"`
```
timbastin: this wont work i am afraid. You need to use the string datatype. The client wont be able to send a JSONB type

internal/core/asset/asset_dto.go
```diff
@@ -75,50 +80,55 @@ type patchRequest struct {
 
 	RepositoryID   *string `json:"repositoryId"`
 	RepositoryName *string `json:"repositoryName"`
+
+	ConfigFiles *database.JSONB `json:"configFiles"`
```
timbastin: Same here

internal/core/asset/asset_dto.go
timbastin: For the create request, we do not even need it at all. Just the patch one

internal/core/compliance/compliance_controller.go
```diff
@@ -56,6 +56,7 @@ func ExtractAttestationPayload(content string) (any, error) {
 // embed the policies in the binary
 //
 //go:embed attestation-compliance-policies/policies/*.rego
+
```
timbastin: Please revert this change. It is a special go:embed directive which should be exactly above the defined variable.

internal/core/org/org_controller.go
```diff
@@ -470,3 +470,11 @@ func (o *httpController) Metrics(ctx core.Context) error {
 	}
 	return ctx.JSON(200, map[string]string{"ownerId": owner})
 }
+
+func (o *httpController) GetConfigFile(ctx core.Context) error {
+	organization := core.GetOrganization(ctx)
+	configID := ctx.Param("config-file")
+
+	configContent := organization.ConfigFiles[configID]
+	return ctx.JSON(200, configContent.(string))
```
timbastin: Please check if the config file exists. If not, return a 404

internal/core/asset/asset_controller.go
```diff
@@ -213,11 +213,27 @@ func (c *httpController) Update(ctx core.Context) error {
...
+		err = a.assetRepository.Update(nil, &asset)
 		if err != nil {
 			return fmt.Errorf("error updating asset: %v", err)
 		}
 	}
 
 	return ctx.JSON(200, asset)
 }
+
+func (a *httpController) GetConfigFile(ctx core.Context) error {
```
timbastin, line 225: Please check if the config  file exists - otherwise return a 404

internal/core/org/org_dto.go
```diff
@@ -36,16 +37,17 @@ type changeRoleRequest struct {
...
+	ContactPhoneNumber     *string        `json:"contactPhoneNumber"`
+	NumberOfEmployees      *int           `json:"numberOfEmployees"`
+	Country                *string        `json:"country"`
+	Industry               *string        `json:"industry"`
+	CriticalInfrastructure bool           `json:"criticalInfrastructure"`
+	ISO27001               bool           `json:"iso27001"`
+	NIST                   bool           `json:"nist"`
+	Grundschutz            bool           `json:"grundschutz"`
+	Description            string         `json:"description"`
+	ConfigFiles            database.JSONB `json:"configFiles"`
```
timbastin: Please change the datatype to string

internal/core/project/project_controller.go
```diff
@@ -412,3 +412,15 @@ func (p *controller) Update(c core.Context) error {
 	}
 	return c.JSON(200, resp)
 }
+
+func (o *controller) GetConfigFile(ctx core.Context) error {
```
timbastin, line 416: Please check if the config  file exists - otherwise return a 404

timbastin (changes requested):

Hubtrick-Git:
now its ready to be reviewed @timbastin

PR #562, opened 2025-04-28
add delete asset method with all assetversions

refoo0:


timbastin (changes requested):
I would prefer overwriting the delete method. Otherwise the delete method exists but just never be used. This makes it hard to maintain

timbastin:
Would be great to have the delete and relation ship deletion done in the same transaction

timbastin (approved):

PR #572, opened 2025-04-30
535 add information about the amount of vulnerabilities with known exploits to the dashboard

Hubtrick-Git:
Added a function for retrieving all CVEs with an existing exploit for a given assetVersion.

### DIFF
--- a/cmd/devguard/api/api.go
+++ b/cmd/devguard/api/api.go
@@ -403,7 +403,7 @@ func BuildRouter(db core.DB) *echo.Echo {
 	componentController := component.NewHTTPController(componentRepository, assetVersionRepository)
 	complianceController := compliance.NewHTTPController(assetVersionRepository)
 
-	statisticsController := statistics.NewHttpController(statisticsService, assetRepository, assetVersionRepository, projectService)
+	statisticsController := statistics.NewHttpController(statisticsService, statisticsRepository, assetRepository, assetVersionRepository, projectService)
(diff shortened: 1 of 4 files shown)


cmd/devguard/api/api.go
```diff
@@ -537,6 +537,7 @@ func BuildRouter(db core.DB) *echo.Echo {
 	assetRouter.GET("/compliance/", complianceController.AssetCompliance)
 	assetRouter.GET("/stats/risk-distribution/", statisticsController.GetAssetVersionRiskDistribution)
 	assetRouter.GET("/stats/cvss-distribution/", statisticsController.GetAssetVersionCvssDistribution)
+	assetRouter.GET("/number-of-exploits/", statisticsController.GetNumberOfExploitableCVES)
```
timbastin: Would be fitting to have this inside /stats/vulns-with-exploits

internal/core/statistics/statistics_asset_controller.go
```diff
@@ -267,3 +269,24 @@ func (c *httpController) getDependencyVulnAggregationStateAndChange(compareTo st
 
 	return c.statisticsService.GetDependencyVulnAggregationStateAndChangeSince(assetVersion.Name, assetVersion.AssetID, calculateChangeTo)
 }
+
+func (c *httpController) GetNumberOfExploitableCVES(ctx core.Context) error {
```
timbastin: I think the method name does not match the functionality anymore. I like the new repository method name. Actually all cves are exploitable - that is basically the definition. The Known Exploits stuff is what differs.

timbastin (changes requested):

timbastin (approved):

PR #641, opened 2025-05-14
add monitoring metrics

refoo0:


internal/core/daemon/deps_dev_daemon.go
```diff
@@ -1,13 +1,20 @@
...
+
 	"github.com/l3montree-dev/devguard/internal/core"
 	"github.com/l3montree-dev/devguard/internal/core/component"
 	"github.com/l3montree-dev/devguard/internal/core/vulndb"
 	"github.com/l3montree-dev/devguard/internal/database/repositories"
+	"github.com/l3montree-dev/devguard/internal/monitoring"
 )
 
 func UpdateDepsDevInformation(db core.DB) error {
+	strat := time.Now()
```
timbastin, line 14:
```suggestion
	start := time.Now()
```

internal/core/daemon/daemon.go
```diff
@@ -56,6 +57,11 @@ func markMirrored(configService config.Service, key string) error {
 }
 
 func Start(db core.DB) {
+	monitoring.DaemonStartedAmount.Inc()
+	start := time.Now()
+	defer func() {
+		monitoring.DaemonStartedDuration.Observe(time.Since(start).Minutes())
```
timbastin: I think this metric should always be around miliseconds right? Since the command only starts a goroutine. Does it add any interesting information?

internal/monitoring/prom.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var DaemonStartedAmount = promauto.NewCounter(prometheus.CounterOpts{
+	Name: "devguard_daemon_started_amount",
+	Help: "The total number of daemon started",
+})
+var DaemonStartedDuration = promauto.NewHistogram(prometheus.HistogramOpts{
```
timbastin: I think those can actually be deleted. It just fires up a goroutine. Not sooo interesting

timbastin (approved):

Issue #768, opened 2025-06-11
Add addtional infos to vex


Hubtrick-Git:
add extra properties to a vex to display how fast members respond to a vulnerability

PR #769, opened 2025-06-11
748 write tests using containers for different functions

Hubtrick-Git:
Test are basically finished for sbom !STILL NEEDS CLEAN UP! 
Still need to test VEX


Hubtrick-Git:
Vex tests will be done in #768

PR #793, opened 2025-06-13
check the admin token first before checking ory kratos

David-Development:


David-Development:
@timbastin was faster I guess.. 🥲

PR #811, opened 2025-06-23
SBOM Report Generation

seb-kw:


internal/core/assetversion/asset_version_controller.go
```diff
@@ -332,3 +341,136 @@ func (a *AssetVersionController) Metrics(ctx core.Context) error {
...
 	})
 }
+
+func (a *AssetVersionController) BuildPDFFromSBOM(ctx core.Context) error {
+	//build the SBOM of this asset version
+	bom, err := a.buildSBOM(ctx)
+	if err != nil {
+		return err
+	}
+	workingDir, err := os.Getwd()
```
timbastin: You will run into issues when multiple users call this endpoint. In general, you should be able to do all of this function in memory only - which will be faster, more reliable and wont have any issues with concurrent requests. It should be even easier to test.

```golang
zipFile := bytes.NewBuffer(nil)
zipWriter := zip.NewWriter(zipFile)
```

Checkout the policy controller how to handle template files which should be embedded inside the binary (line 264 policy_controller.go) // go:embed directive. https://towardsdev.com/simplify-your-go-projects-with-embed-fs-a-practical-guide-fc827590d879

internal/core/assetversion/asset_version_service.go
```diff
@@ -881,3 +882,91 @@ func getDatesForVulnerabilityEvent(vulnEvents []models.VulnEvent) (time.Time, ti
 
 	return firstIssued, lastUpdated, firstResponded
 }
+
+func markdownTableFromSBOM(bom *cdx.BOM) string {
```
timbastin: I think this would benefit a lot from golang templating: https://pkg.go.dev/text/template

Since we are currently in the process of rewriting such functions which are using the string builder to use the standard library template package, it would be great to have this function correct, right from the beginning

internal/core/assetversion/asset_version_service.go
```diff
@@ -881,3 +882,91 @@ func getDatesForVulnerabilityEvent(vulnEvents []models.VulnEvent) (time.Time, ti
...
+				}
+			}
+		}
+		tableRow := "| " + component.BOMRef + " | " + component.Name + " | " + component.Version + " | " + licenseString + " | " + string(component.Type) + " |\n"
+		markdownText.WriteString(tableRow)
+	}
+	return markdownText.String()
+}
+
+func createYAMLMetadata(orgName string, projectName string, assetVersionName string) string {
```
timbastin: you can just convert a struct to yaml:  https://dev.to/sagartrimukhe/generate-yaml-files-in-golang-29h1. This basically works the same like the json marshalling.

```golang
yaml.Marshal(...)
```

internal/core/assetversion/asset_version_service.go
```diff
@@ -881,3 +882,91 @@ func getDatesForVulnerabilityEvent(vulnEvents []models.VulnEvent) (time.Time, ti
...
+	yamlText.WriteString(fmt.Sprintf("  generation_date: %s. %s %s\n", strconv.Itoa(today.Day()), today.Month().String(), strconv.Itoa(today.Year())))
+	yamlText.WriteString(title1)
+	yamlText.WriteString(title2)
+	yamlText.WriteString(fmt.Sprintf("  organization_name: %s\n", orgName))
+	// TO-DO: add sha hash to test the integrity
+	yamlText.WriteString("  integrity: sha265:3d8ce29bd449af3709535e12a93e0 fa2cea666912c3d37cf316369613533888d\n")
+	return yamlText.String()
+}
+
+func createTitlesFromProjectName(projectName string) (string, string) {
```
timbastin: this function looks odd. What exactly are you tring todo? Maybe you can give some input and desired output examples.

internal/core/assetversion/asset_version_service_test.go
```diff
@@ -81,3 +90,95 @@ func TestDiffScanResults(t *testing.T) {
 		assert.Empty(t, notDetectedByCurrentScannerAnymore)
 	})
 }
+
+func TestFileCreationForPDFSBOM(t *testing.T) {
+	t.Run("test the new functions", func(t *testing.T) {
```
timbastin: The name of the test should give a hint about what you are really expecting of the test.

Like: `t.Run("should generate a matching markdown table from an sbom")`

I think the issue you had here in finding a fitting test name hints you to the real problem: You are doing way to many in a single test. A single t.Run should verify a *single branch* in the function you are testing. Thus, each if statement creates another branch. So for each if, you would need at least another t.Run statement.

A `TestXYZ` function should test a single function like `markdownTableFromSBOM ` in *isolation*. 

What you are doing here is an end to end test - even calling the real api (which is an antipattern).

internal/core/assetversion/asset_version_service_test.go
```diff
@@ -81,3 +90,95 @@ func TestDiffScanResults(t *testing.T) {
...
+		bom := cdx.BOM{
+			Components: &[]cdx.Component{
+				{BOMRef: "pkg:deb/debian/gcc-12@12.2.0", Name: "debian/gcc-12", Version: "12.2.0-14", Type: "application", Licenses: &cdx.Licenses{cdx.LicenseChoice{License: &cdx.License{ID: "Apache-2.0"}}}},
+				{BOMRef: "pkg:deb/debian/libc6@2.36-9+deb12u10", Name: "debian/libc6", Version: "2.36-9+deb12u10", Type: "library", Licenses: &cdx.Licenses{cdx.LicenseChoice{License: &cdx.License{ID: "MIT"}}}},
+				{BOMRef: "pkg:deb/debian/libstdc++6@12.2.0-14", Name: "debian/libstdc++6", Version: "12.2.0-14", Type: "library", Licenses: &cdx.Licenses{}},
+			},
+		}
+
+		markdownTable := markdownTableFromSBOM(&bom)
```
timbastin: Do some assertions on the created string and finish your first test.

internal/core/assetversion/asset_version_service_test.go
```diff
@@ -81,3 +90,95 @@ func TestDiffScanResults(t *testing.T) {
...
+			slog.Error("URL of the pdf api is missing")
+			t.Fail()
+		}
+		fmt.Printf("URL = %s", pdfAPIURL)
+		req, err := http.NewRequest("POST", pdfAPIURL, &buf)
+		assert.Nil(t, err)
+		req.Header.Set("Content-Type", mpw.FormDataContentType())
+		client := &http.Client{}
+
+		resp, err := client.Do(req)
```
timbastin: That http request should be mocked. https://golang.cafe/blog/golang-httptest-example.html

report-templates/sbom/markdown/sbom.md
```diff
@@ -0,0 +1,7 @@
+# SBOM {.unlisted .unnumbered}
```
timbastin:
```suggestion
# SBOM
```

report-templates/sbom/template/template.tex
```diff
@@ -0,0 +1,210 @@
...
+\usepackage{xcolor}
+\usepackage{tikz}
+\usepackage{microtype}
+\usepackage[absolute,overlay]{textpos}
+\usepackage{tocloft}
+% Gets corrected using \newgeometry{...} in the document
+% Used for the cover page
+\geometry{top=0pt, left=0pt, right=0pt, bottom=0pt}
+
+\interfootnotelinepenalty=10000
```
timbastin: I think we are not using any footnotes in the sbom

internal/core/assetversion/asset_version_service_test.go
Hubtrick-Git: This test was not supposed to be for real unit testing but more a test to see whether the function works. I did not intend to have this test merged with main. After improving the code I had in mind to write a proper test function for this.

timbastin: ❤️

PR #820, opened 2025-06-30
Feautre/jira integration

refoo0:


internal/core/integrations/jira/jira_client.go
```diff
@@ -0,0 +1,304 @@
...
+
+	var projects []*Project
+	if err := json.NewDecoder(resp.Body).Decode(&projects); err != nil {
+		return nil, fmt.Errorf("failed to decode projects response: %w", err)
+	}
+
+	return projects, nil
+}
+
+func jiraRequest(client Client, method string, url string, body io.Reader) (*http.Response, error) {
```
timbastin:
```suggestion
func (client Client) jiraRequest(method string, url string, body io.Reader) (*http.Response, error) {
```

internal/core/integrations/jira/jira_client.go
```diff
@@ -0,0 +1,304 @@
...
+
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
+
+	req.SetBasicAuth(client.UserEmail, client.AccessToken)
+	return http.DefaultClient.Do(req)
+
+}
+
+func ParseWebHook(payload []byte) (*WebhookEvent, error) {
```
timbastin:
```suggestion
func ParseWebhook(payload []byte) (*WebhookEvent, error) {
```

internal/core/integrations/jira/jira_client.go
timbastin: You can handle any error right here. Read the body into a bytes.Buffer - then you can detect errors. After detection, set the response.Body to your bytes buffer and call Seek(0) to move the current reading pointer to the start.

timbastin (approved):

PR #822, opened 2025-07-01
fix first-party-vulns gitlab-webhook-events handling

refoo0:


internal/core/integrations/gitlabint/gitlab_webhook.go
```diff
@@ -231,7 +232,7 @@ func (g *GitlabIntegration) HandleWebhook(ctx core.Context) error {
 
 		}
 	}
-
+	fmt.Println("projectID", projectID, "issueID", issueID, "vulnEvent", vulnEvent)
```
timbastin: This can be deleted 🙂

PR #824, opened 2025-07-02
sbom are now validated by the cdx validater

Hubtrick-Git:
changed the sbom.json generation to be sound with the cdx standard
working on the sbom.xml


Hubtrick-Git:
Both formats are now validated by https://cyclonedx.github.io/cyclonedx-web-tool/validate
XML: missing name space
Licenses : 
    cannot have an empty License Choice
    cannot have id AND Name
    non-official licenses, like non-standard need to be in the Name field

timbastin (approved):

PR #866, opened 2025-07-08
implement pagination for project listing and add search functionality

refoo0:


internal/core/project/project_service.go
```diff
@@ -215,6 +254,73 @@ func (s *service) ListAllowedProjects(c core.Context) ([]models.Project, error)
...
+}
+
+func (s *service) ListAllowedProjectsPaged(c core.Context) (core.Paged[models.Project], error) {
+
+	pageInfo := core.GetPageInfo(c)
+	search := c.QueryParam("search")
+
+	// get all projects the user has at least read access to
+	rbac := core.GetRBAC(c)
+	projectSliceOrProjectIDSlice, err := rbac.GetAllProjectsForUser(core.GetSession(c).GetUserID())
```
timbastin, line 267: It would be much more efficient to forward any filtering parameters to this function. Otherwise you are fetching a huge amount of projects (maybe 200) and filtering those down to 10. Especially since we are expecting to talk to an external system, which might be slow, it would be reasonable to send the filter values as well.

timbastin: Same when requesting projects from our own database.

PR #874, opened 2025-07-10
update ListAllowedProjects to refresh project list

refoo0:


timbastin (approved):

internal/core/project/project_service.go
```diff
@@ -183,7 +183,12 @@ func (s *service) ListAllowedProjects(c core.Context) ([]models.Project, error)
 			slog.Info("enabled community managed policies for project", "projectSlug", project.Slug, "projectID", project.ID)
 		}
 
-		return slice, err
```
timbastin, line 186: Return created and updated right here. Updated is currently not assigned as return value from upsert split. I am trying to avoid the extra db call

timbastin (changes requested):

timbastin (approved):

PR #878, opened 2025-07-12
844 license risks

Hubtrick-Git:
Added license risk table to the database
Dependency Vuln Scans now automatically check each dependency if they have a valid license
Added endpoints to interact with License Risks just like dependency-/firstParty-Vulns
On-top a user can manually set a final license


Hubtrick-Git:
Do we also need to include the scanner ids in the Calculate hash function since, the super key is technically assetID, assetVersionName, pURL and scanner ID since the same component in the same assetVersion can be found by multiple scanner leading to duplicate vulnIDs?

cmd/devguard/api/api.go
```diff
@@ -679,6 +676,14 @@ func BuildRouter(db core.DB) *echo.Echo {
 	firstPartyVulnRouter.POST("/:firstPartyVulnID/mitigate/", firstPartyVulnController.Mitigate, neededScope([]string{"manage"}), projectScopedRBAC(core.ObjectAsset, core.ActionUpdate))
 	firstPartyVulnRouter.GET("/:firstPartyVulnID/events/", vulnEventController.ReadAssetEventsByVulnID)
 
+	assetVersionRouter.POST("/license-risk/", licenseRiskController.Create)
```
```suggestion
	assetVersionRouter.POST("/license-risks/", licenseRiskController.Create)
```

Hubtrick-Git:
did not yet merge with main, makes the code review probably easier and there is yet again a problem with go embed in the asset version controller

timbastin (approved):

PR #883, opened 2025-07-16
828 external entity providers there are projects missing in the org overview

Hubtrick-Git:
Fixed a nil pointer exception and now all groups are fetched successfully

### DIFF
--- a/internal/core/integrations/gitlabint/gitlab_integration.go
+++ b/internal/core/integrations/gitlabint/gitlab_integration.go
@@ -336,7 +336,9 @@ func (g *GitlabIntegration) ListGroups(ctx core.Context, userID string, provider
 	}
 	// get the groups for this user
 	groups, _, err := gitlabClient.ListGroups(ctx.Request().Context(), &gitlab.ListGroupsOptions{
-		ListOptions: gitlab.ListOptions{PerPage: 100}, // only list groups where the user has at least owner permissions
+		ListOptions: gitlab.ListOptions{PerPage: 100},
+		//MinAccessLevel: utils.Ptr(gitlab.ReporterPermissions),
+		// only list groups where the user has at least reporter permissions
 	})
 
 	if err != nil {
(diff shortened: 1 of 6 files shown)


timbastin (approved):

internal/core/integrations/gitlabint/gitlab_integration.go
```diff
@@ -336,15 +336,36 @@ func (g *GitlabIntegration) ListGroups(ctx core.Context, userID string, provider
...
 
 	if err != nil {
 		slog.Error("failed to list groups", "err", err)
 		return nil, err
 	}
 
-	return utils.Map(groups, func(el *gitlab.Group) models.Project {
+	cleanedGroups := make([]*gitlab.Group, 0, len(groups))
+
+	for _, group := range groups {
```
timbastin, line 350: you should be able todo this in parallel. Try to spawn some goroutines to speed things up. You can do syncronization using a mutex.

timbastin (changes requested):

internal/accesscontrol/external_entity_provider_rbac_test.go
```diff
@@ -175,7 +176,7 @@ func TestHasAccess(t *testing.T) {
 			nil,
 			thirdpartyIntegrationMock,
 			"external-entity-provider-id",
-			"",
+			utils.Ptr(""),
```
timbastin: I think providing nil here fits better

internal/accesscontrol/external_entity_provider_rbac_test.go
```diff
@@ -190,7 +191,7 @@ func TestHasAccess(t *testing.T) {
...
```
timbastin: Same here

internal/core/integrations/gitlabint/gitlab_oauth2.go
```diff
@@ -106,7 +106,7 @@ func parseGitlabEnvs() map[string]gitlabEnvConfig {
 				conf.botUserAccessToken = value
 			case "admintoken":
 				if value == "" {
-					conf.adminToken = nil
+					conf.adminToken = utils.Ptr("")
```
timbastin: Nil here was fine actually. Like there is no admin token configured

timbastin (changes requested):

PR #919, opened 2025-07-18
VeX Upload

timbastin:


internal/core/assetversion/asset_version_controller.go
```diff
@@ -444,7 +444,6 @@ func (a *AssetVersionController) BuildPDFFromSBOM(ctx core.Context) error {
 	return err
 }
 
-//go:embed report-templates/*
```
timbastin, line 447: this needs to be reverted.

internal/core/common_interfaces.go
```diff
@@ -147,6 +147,7 @@ type DependencyVulnRepository interface {
 	ApplyAndSave(tx DB, dependencyVuln *models.DependencyVuln, vulnEvent *models.VulnEvent) error
 	GetDependencyVulnsByDefaultAssetVersion(tx DB, assetID uuid.UUID, scannerID string) ([]models.DependencyVuln, error)
 	ListUnfixedByAssetAndAssetVersionAndScannerID(assetVersionName string, assetID uuid.UUID, scannerID string) ([]models.DependencyVuln, error)
+	GetAllVulnsForAssetVersion(tx DB, assetID uuid.UUID, assetVersionName string) ([]models.DependencyVuln, error)
```
timbastin, line 150: I think this function will do the same if you provide an empty scanner id:

`GetDependencyVulnsByAssetVersion`

PR #921, opened 2025-07-18
Draft: add vulnerability template

David-Development:
[Example PDF](https://github.com/user-attachments/files/21312709/Archiv.3.pdf)


timbastin:
This is already done! Thanks a lot!

PR #923, opened 2025-07-18
Update asset name and artifact suffix in devguard-scanner workflow

refoo0:


.github/workflows/devguard-scanner.yaml
```diff
@@ -83,158 +47,81 @@ jobs:
...
+      build-args: "--context=. --dockerfile=Dockerfile --build-arg GITHUB_REF_NAME=$GITHUB_REF_NAME"       
+
 
   build-scanner-image:
     uses: l3montree-dev/devguard-action/.github/workflows/build-image.yml@main
     with:
-      image-suffix: "-scanner"
-      artifact-suffix: "-scanner"
-      asset-name: l3montree-cybersecurity/projects/devguard/assets/devguard-scanner
+      artifact-name: "-scanner"
```
timbastin: This way the other docker image would be call -scanner right? Maybe remove the leading dash

timbastin (changes requested):

timbastin (approved):

PR #929, opened 2025-07-21
removed devguard.org link and moved to frontend to make it clickable

5byuri:
https://github.com/l3montree-dev/devguard/issues/928

### DIFF
--- a/.kratos/identity.schema.json
+++ b/.kratos/identity.schema.json
@@ -53,7 +53,7 @@
                 },
                 "confirmedTerms": {
                     "type": "boolean",
-                    "title": "I agree to the terms of use (devguard.org/terms-of-use)",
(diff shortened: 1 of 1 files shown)


5byuri:
image
now the TOS is clickable

5byuri:
Yes! 
Ory uses the change-identity.schema, as of right now we have `devguard.org/terms-of-use` actually in the title.
https://github.com/l3montree-dev/devguard-web/blob/main/src/components/kratos/NodeInputCheckbox.tsx
I moved the link to the frontend and made it clickable with an anchor.

5byuri:
if we dont change the json, it would look like this:
image

seb-kw (approved):

PR #954, opened 2025-07-31
Added hints about other instances of vulns when getting

Hubtrick-Git:


internal/core/compliance/attestation-compliance-policies
```diff
@@ -1 +1 @@
-Subproject commit 11781c0726fb7e8ab7181fc0f4309e8642bd97e6
```
timbastin: please call `git submodule update`

PR #998, opened 2025-08-11
Fixed closing of tickets after fixing first party vulns

Hubtrick-Git:
After handling first party vuln results we now also update all tickets associated with the handled first party vulns, especially closing the fixed ones.


internal/database/repositories/first_party_vuln_repository.go
```diff
@@ -19,11 +19,20 @@ func NewFirstPartyVulnerabilityRepository(db core.DB) *firstPartyVulnerabilityRe
 }
 
 func (repository *firstPartyVulnerabilityRepository) ListByScanner(assetVersionName string, assetID uuid.UUID, scannerID string) ([]models.FirstPartyVuln, error) {
+	// tx core.DB missing (or chosen not to be implemented) ?
```
Hubtrick-Git, line 22: Do we want to also add the option to put this in a parent transaction ?

PR #1004, opened 2025-08-13
997 integrate debian license scanning

Hubtrick-Git:
If a component is located in the debian namespace we now check metadata.ftp-master.debian to get the license informationen and then use google/licensecheck to determine the license from the copyright information.


internal/core/component/component_service.go
```diff
@@ -75,80 +80,97 @@ func (s *service) RefreshComponentProjectInformation(project models.ComponentPro
...
 			}
+		} else {
+			slog.Warn("no related projects found", "purl", pURL)
 		}
-	} else {
-		slog.Warn("no related projects found", "purl", pURL)
 	}
-
 	return component, nil
 }
```
Hubtrick-Git, line 175: In the GetLicense function we always return nil, never an error maybe we can remove the error as return value entirely ?

timbastin: Good idea! nevertheless, i will merge it now to get the new licenses into production

PR #1289, opened 2025-10-19
Fix scanner to return 400 Bad Request for invalid SBOM uploads

### LINKED ISSUE #1278: Scanner fails without error message when uploading an invalid sbom (using sbom command)
The backend fails with 500 which should actually be a 400. 

2:56PM ERR echohttp/server.go:30 json: cannot unmarshal array into Go struct field Component.metadata.component.version of type string method=POST path=http://api.devguard.opencode.de/api/v1/scan/

- [ ] Fail with 400
- [ ] Print server provided error message in sbom command

mine-13-zoom:
This fixes issue #1278 where the scanner was returning 500 errors for invalid SBOMs instead of 400. Now it properly sends a 400 Bad Request with a clear message.

## What changed
- In scan_controller.go, wrapped the CycloneDX decoding errors in echo.NewHTTPError(400, "Invalid SBOM format") for both ScanDependencyVulnFromProject and ScanSbomFile methods.
- This matches how other similar errors are handled in the codebase, like in UploadVEX.

## Why this matters
- Users uploading bad SBOMs will now get a proper 400 error instead of a 500, and the client prints the server message so they know what went wrong.
- No breaking changes, just better error handling.

Closes #1278

### DIFF
--- a/internal/core/vulndb/scan/scan_controller.go
+++ b/internal/core/vulndb/scan/scan_controller.go
@@ -364,7 +364,7 @@ func (s *HTTPController) ScanDependencyVulnFromProject(c core.Context) error {
 	decoder := cdx.NewBOMDecoder(c.Request().Body, cdx.BOMFileFormatJSON)
 	defer c.Request().Body.Close()
 	if err := decoder.Decode(bom); err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


timbastin (approved):
Awesome! Thanks a lot!

PR #1332, opened 2025-10-30
feat: add --outputPath flag to scanning commands

### LINKED ISSUE #1302: Add flag --outputPath to scanning commands to save sarif.json files (e.g. to upload as artifact)
It would be great when devguard-scanner would support the flag `--outputPath`. This would enable e.g .the ci components to set a stable and user defined path to save the result files.

mine-13-zoom:
## Summary
Add --outputPath flag to scanning commands (sast, secret-scanning, iac, sarif) to enable saving SARIF reports to user-defined paths for CI/CD artifact uploads.

## Changes
- Add --outputPath flag to all scanning commands
- Update scan functions to accept and use output path parameter
- Create output directories if they don't exist
- Preserve original behavior when outputPath is not specified
- Update help examples to show new flag usage

## Fixes
Resolves #1302

### DIFF
--- a/cmd/devguard-scanner/commands/iac.go
+++ b/cmd/devguard-scanner/commands/iac.go
@@ -67,6 +67,11 @@ func iacScan(p, outputPath string) (*common.SarifResult, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, "could not remove file")
 		}
+	} else {
+		err = os.Rename(sarifFilePath, outputPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not move file to output path")
+		}
 	}
 
 	return &sarifScan, nil


cmd/devguard-scanner/commands/secret_scanning.go
```diff
@@ -30,15 +30,17 @@ You may pass the target as the first positional argument instead of using
...
 
 	scanner.AddFirstPartyVulnsScanFlags(secretScanningCommand)
+	secretScanningCommand.Flags().String("outputPath", "", "Path to save the SARIF report. If not specified, the report will only be uploaded to DevGuard.")
 	return secretScanningCommand
 }
 
-func secretScan(p string) (*common.SarifResult, error) {
+func secretScan(p, outputPath string) (*common.SarifResult, error) {
 	dir := os.TempDir()
 	dir = path.Join(dir, "secret-scanning")
```
timbastin, line 44: I think we can move those "directory" definition and the creation in line 48 into your else branch. We wont need to create those if an output path is provided right?

cmd/devguard-scanner/commands/secret_scanning.go
```diff
@@ -30,15 +30,17 @@ You may pass the target as the first positional argument instead of using
 Example:
 	devguard-scanner secret-scanning --path ./my-repo
 	devguard-scanner secret-scanning ./my-repo
+	devguard-scanner secret-scanning --path ./my-repo --outputPath secrets.sarif.json
 `,
 		RunE: sarifCommandFactory("secret-scanning"),
 	}
 
 	scanner.AddFirstPartyVulnsScanFlags(secretScanningCommand)
+	secretScanningCommand.Flags().String("outputPath", "", "Path to save the SARIF report. If not specified, the report will only be uploaded to DevGuard.")
```
timbastin: Lets promote this flag into the `scanner.AddFirstPartyVulnsScanFlags` function. What do you think? You added it (correctly) to all first party scans. Just some deduplication

cmd/devguard-scanner/commands/sast.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func sastScan(p string) (*common.SarifResult, error) {
+func sastScan(p, outputPath string) (*common.SarifResult, error) {
 	dir := os.TempDir()
 	dir = path.Join(dir, "sast")
```
timbastin, line 19: I think we can move this into the else branch as well. Exactly like in secret and iac scan

cmd/devguard-scanner/commands/sarif.go
```diff
@@ -232,10 +247,15 @@ func sarifCommandFactory(scannerID string) func(cmd *cobra.Command, args []strin
...
+		outputPath, err := cmd.Flags().GetString("outputPath")
+		if err != nil {
+			return errors.Wrap(err, "could not get outputPath flag")
+		}
+
 		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
 		defer cancel()
 
-		sarifResult, err := executeCodeScan(scannerID, config.RuntimeBaseConfig.Path)
+		sarifResult, err := executeCodeScan(scannerID, config.RuntimeBaseConfig.Path, outputPath)
```
timbastin: I think we can promote your new config flag to the config.RuntimeBaseConfig. Maybe you can check it out. It should be as simple as giving the struct another property `OutputPath`. The flag parsing does already happen in the pre run functions of each command or the root command.

timbastin (commented):
This looks great! I added a few suggestions. What do you think about it?

cmd/devguard-scanner/commands/iac.go
```diff
@@ -52,9 +62,11 @@ func iacScan(p string) (*common.SarifResult, error) {
...
 	// remove the file
-	err = os.Remove(path.Join(dir, "results_sarif.sarif"))
-	if err != nil {
-		return nil, errors.Wrap(err, "could not remove file")
+	if outputPath == "" {
+		err = os.Remove(sarifFilePath)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not remove file")
+		}
 	}
```
timbastin, line 75: Everything looks great! Just a really small suggestion which I did not notice earlier - sorry for that:

Currently the iac result is not saved to the output path like in the other commands but to Dir(outputFilePath) + results_sarif.sarif since we can only provide the output directory to checkov and it will always save it in this result file. What do you think about moving the sarifFilePath to the output path after reading if defined? That would be a great UX since all commands behave exactly the same. Just adding a else branch right here and moving the file should be enough. Do you have thoughts on this?

mine-13-zoom: Ok yeah, I get it, c862dbb6f0501ca54f179cfe7bfe513dd337a252 should account for this.

PR #1361, opened 2025-11-04
changed sbom generation to be sound with the spdx standard

Hubtrick-Git:
SBOM was not sound with SPDX V1.6 standard due to:
- Metadata/Component/Type not set -> now its "application"
- License not properly handled:
      - if we have an unknown or non-standard license write them into the name attribute instead of id attribute
      - if we have a license expression ( containing AND,OR,WITH), write the expression into the expression attribute
      

Open Questions:
Metadata/Component/Type is now set in normalize.FromNormalizedCdxBom which is also used by VEX.json.
Does it make sense to change the value in the normalize function or rather in the calling function AND
is application the fitting type here?

List of valid values for Metadata/Component/Type:
image


timbastin:
I think application is pretty generic and basically fits every SBOM. Currently I do not know any requirement about the type attribute. I think you made a good choice. Let's stick with application until someone says we shouldn't

PR #1400, opened 2025-11-13
implement DeleteEventByID functionality in vulnEventController and repository

refoo0:


internal/core/events/vuln_event_controller.go
```diff
@@ -109,3 +109,26 @@ func (c vulnEventController) ReadEventsByAssetIDAndAssetVersionName(ctx core.Con
...
 }
+
+func (c vulnEventController) DeleteEventByID(ctx core.Context) error {
+	eventID := ctx.Param("eventID")
+	if eventID == "" {
+		return echo.NewHTTPError(400, "eventID is required")
+	}
+
+	asset := core.GetAsset(ctx)
+	hasAccess, err := c.vulnEventRepository.HasAccessToEvent(asset.ID, eventID)
```
timbastin: I think there is no need for this check, since you registered it as a subroute of the dependencyvulnrouter. There is permission checking already done inside the middlewares

timbastin (changes requested):
Added a comment :)

PR #1470, opened 2025-12-05
Kyverno2sarif and sarif2md

5byuri:
ported code from opencode to devguard


cmd/devguard-scanner/commands/kyverno2sarif.go
```diff
@@ -0,0 +1,256 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+type kyvernoTestResult struct {
```
timbastin, line 14: I think we can reuse a lot of oour sarif_result_dto.go types right here. Checkout this files

PR #1479, opened 2025-12-08
License Risk Tickets

Hubtrick-Git:
For some reason up until now the backend had no functionality for handling create Issue requests for license risks, yet the button was developed in the front end.
 - Now we can successfully open tickets using the button
 - Tickets include the vulnerable component as well as links to OSI and DevGuard

!Not tested for Github integration!

The code may not be flawless...

### DIFF
--- a/integrations/githubint/github_integration.go
+++ b/integrations/githubint/github_integration.go
@@ -929,8 +929,8 @@ func (githubIntegration *GithubIntegration) createLicenseRiskIssue(ctx context.C
 		Color:       github.String("182654"),
 	})
 	if err != nil {
+		// nevertheless try to create the comment, thus not returning here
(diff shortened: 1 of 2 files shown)


integrations/githubint/github_integration.go
```diff
@@ -935,6 +979,7 @@ func (githubIntegration *GithubIntegration) createDependencyVulnIssue(ctx contex
 
 		if err != nil {
 			slog.Error("could not update label", "err", err)
+			return nil, err
```
Hubtrick-Git: Is there are reason these were missing?

timbastin: because we want to continue even if the label editing fails. I will remove your return statements. Besides that the PR looks fine!

timbastin (approved):

PR #1480, opened 2025-12-09
1467 devguard scanner automated attestation

timbastin:


compliance/rego.go
```diff
@@ -128,6 +130,8 @@ func NewPolicy(filename string, content string) (*PolicyFS, error) {
 	}, nil
 }
 
+// here, output should be sarif from now
```
timbastin: The output of this function should not be sarif. Maybe the PolicyEvaluation struct would benefit from a toSarif function

PR #1482, opened 2025-12-09
Add ecosystem to check affected components

refoo0:


normalize/purl.go
```diff
@@ -30,9 +30,6 @@ func normalizePurl(purl string) string {
 	if err != nil {
 		return purl
 	}
-
-	// remove any query parameters
-	purl = strings.Split(purl, "?")[0]
```
timbastin, line 35: We should really normalize the purl right here and put it into the canonical form using the packageurl package

services/asset_version_service.go
```diff
@@ -491,6 +491,92 @@ func diffVulnsBetweenBranches[T Diffable](foundVulnerabilities []T, existingVuln
...
+			}
+
+		}
+	}
+
+	if len(vulnsToUpdate) == 0 {
+		return existingVulns, existingVulnsOnOtherBranch, nil
+	}
+
+	db, err := shared.DatabaseFactory()
```
timbastin: I would prefer the GetDB method from any repository

normalize/purl.go
```diff
@@ -25,14 +25,17 @@ func BeautifyPURL(pURL string) (string, error) {
 
 // returns the normalized purl AND the component type
 func normalizePurl(purl string) string {
-	// unescape the purl
-	purl, err := url.PathUnescape(purl)
+
+	parsedPurl, err := packageurl.FromString(purl)
```
timbastin, line 29:
```suggestion
	parsedPurl, err := packageurl.FromString(url.PathUnescape(purl))
```

I think we need todo it like this

services/asset_version_service.go
```diff
@@ -503,6 +585,13 @@ func (s *assetVersionService) handleScanResult(userID string, artifactName strin
 		slog.Error("could not get existing dependencyVulns on default branch", "err", err)
 		return []models.DependencyVuln{}, []models.DependencyVuln{}, []models.DependencyVuln{}, err
 	}
+
```
timbastin, line 588:
```suggestion
// this is just for migration.
// the call can be removed after all assets were scanned again
```

vulndb/scan/purl_comparer.go
```diff
@@ -57,26 +61,57 @@ func (comparer *PurlComparer) GetAffectedComponents(purl, version string) ([]mod
 	// Step 4: Create search key (purl without version)
 	parsedPurl.Version = ""
 	searchPurl := parsedPurl.ToString()
+	searchPurl = strings.Split(searchPurl, "?")[0]
```
timbastin: lets not use Split here. You should be able to get the string without qualifiers straight from the parsed purl

timbastin (changes requested):

PR #1505, opened 2025-12-16
sbom to pdf crashed becuase of a nil pointer, instead of now giving t…

5byuri:
…he licenses via the pointer we are now using string slices, if there is no license or string then we will set that to unknown, now len.license will always have a valide slice to point to and not a potential nil-pointer anymore


timbastin (approved):
Wow great job @5byuri

PR #1635, opened 2026-02-03
fix return errors in ListRepositories method

refoo0:


integrations/gitlabint/gitlab_client.go
```diff
@@ -94,9 +95,8 @@ func (gitlabOrgClient *gitlabBatchClient) ListRepositories(search string) ([]git
 		wg.Go(func() ([]gitlabRepository, error) {
 			result, _, err := client.ListProjects(context.TODO(), options)
 			if err != nil {
-				return nil, err
+				slog.Warn("failed to list projects from gitlab client", "err", err, "clientID", client.GetClientID())
```
timbastin, line 98:
```suggestion
				slog.Warn("failed to list projects from gitlab client", "err", err, "clientID", client.GetClientID())
				return nil, nil
```

PR #1769, opened 2026-03-02
Instance admin feature

seb-kw:
Ensure CleanupOrphanedRecordsSQL is also callable


seb-kw:
Ensure that setting an admin for a sso org is covered

seb-kw:
Watch out because of badge api as we would remove the admin token

seb-kw:
Ensure to add a feature to change asset slugs

seb-kw:
We also need an admin authed endpoint that confirms a given org exists and return the owner email address of that org. 

We additionally need an admin authed endpoint that returns for an owner all other orgs this owner owns.

Hubtrick-Git:
maybe we should also be able to reset the admin session timer somehow

Hubtrick-Git:
- [x] Manage Admins (Add, revoke, list)
- [X] Change Asset Slugs manually
- [X] Get owner information about org (if exists)
- [X] Get orgs with owner role for user
- [X] Usage statistics

## TODO
- [x] Disable Org Creation
- [x] Instance Statistics (Vulns, Components, etc)
    - [x] Average Remediation time for vulns inside org
    - [x] Vuln Classification (total amount per severity)
    - [x] Most vulnerable projects
    - [x] Most vulnerable assets
    - [x] Most vulnerable artifacts
- [x] Cache Statistics !!!

## TBD 
- [ ] Database dump (sanitized)

Hubtrick-Git:
@seb-kw currently we pass a limit to the backend when we want e.g. the ecosystem package distribution inside an org. What about if we change the limit to a threshold (relative or absolute?). Instead of "give me the top 10", give me the top 95% or at least  1k components . What do you think about that?

accesscontrol/casbin_rbac.go
```diff
@@ -72,6 +72,22 @@ func (c *casbinRBAC) GetOwnerOfOrganization() (string, error) {
 	return strings.TrimPrefix(listOfUsers[0], "user::"), nil
 }
 
+func (c *casbinRBAC) GetAdminsOfOrganization() ([]string, error) {
+	concurrencyMutex.Lock()
+	implicitUsers, err := c.enforcer.GetImplicitUsersForRole("role::admin", "domain::"+c.domain)
+	if err != nil {
+		return nil, err
+	}
+	concurrencyMutex.Unlock()
```
timbastin: We need to defer the unlock. Otherwise this would lock forever if an error happens

Hubtrick-Git: sorry thats a huge overlook, I will fix the mutex locking patterns rightaway.

seb-kw:
Screenshot 2026-06-08 at 17 55 16

- [x] 1.5 for Avg. Open Vulnerabilities per Organisation seems strange
- [x] 170 Projects with ticket sync enabled = 170 Projects with a GitLab integration is wrong. Here 0 have ticket sync enabled.

Hubtrick-Git:
The 170 tickets sync currently  mirrors the 170 gitlab integrations. The 170 gitlab integrations probably stem from the 170 open code projects imported via the SSO.

timbastin (approved):

PR #1781, opened 2026-03-10
Crowdsourced vexing

timbastin:


crowdsourcevexing/crowdsourced_vexing.go
```diff
@@ -0,0 +1,374 @@
...
+// The sum of all factors per user converges to 1 and never exceeds it.
+// This prevents a single user from having too much influence by creating many organizations/projects and voting multiple times.
+func (t *userVoteTracker) recordVoteAndGetFactor(organization Organization) float64 {
+	creator := organization.CreatedBy
+	priorVotes := t.voteCounts[creator]
+	t.voteCounts[creator]++
+	return float64(priorVotes + 1)
+}
+
+func PathPatternMatchesPath(inPath, inPattern []string) bool {
```
timbastin: You dont need to export this function right here.


```suggestion
func pathPatternMatchesPath(inPath, inPattern []string) bool {
```

The path `[A, B, C]` and the pattern `[*, C, X]` will match. Your `endIndex` will be `2` and the `inPattern[0]` == `*`

Maybe you can checkout `vex_rule_dto.go`. We have this algorithm right there. https://github.com/l3montree-dev/devguard/blob/main/dtos/vex_rule_dto.go#L54

crowdsourcevexing/crowdsourced_vexing.go
```diff
@@ -0,0 +1,374 @@
...
+				}
+			}
+		}
+		return false
+	}
+}
+
+// If the pattern matches with the path, return the length of how far apart the two critical
+// dependecies are (accounts for wildcards)
+func PathPatternMatchLength(inPath, inPattern []string) (int, error) {
```
timbastin: same right here:


```suggestion
func pathPatternMatchLength(inPath, inPattern []string) (int, error) {
```

crowdsourcevexing/crowdsourced_vexing.go
```diff
@@ -0,0 +1,374 @@
...
+	// Filtering for VexRules that apply to the dependecy tree
+	// Deduplucate VexRules based on organizationn and project to avoid replay
+	// (every combination of organization and project will be allow to have one non-contradicting VexRule for a Path submitted)
+	for _, rule := range inVexRules {
+
+		// For each VexRule, find organization and project id
+		rulePath := PathToString(rule)
+
+		var asset Asset
+		for _, asst := range inAssets {
```
timbastin: You can build a map for this before the loop. Then you would avoid the nested for

crowdsourcevexing/crowdsourced_vexing.go
```diff
@@ -0,0 +1,374 @@
...
+				asset = asst
+				break
+			}
+		}
+		if asset.ID == "" {
+			slog.Error("failed to find asset for VEX rule", "assetID", rule.AssetID)
+			continue
+		}
+		var project Project
+		for _, proj := range inProjects {
```
timbastin: This would be more readable by building a map up front as well.

crowdsourcevexing/crowdsourced_vexing.go
```diff
@@ -0,0 +1,374 @@
...
+				project = proj
+				break
+			}
+		}
+		if project.ID == "" {
+			slog.Error("failed to find project for asset", "projectID", asset.ProjectID)
+			continue
+		}
+		var organization Organization
+		for _, org := range inOrganizations {
```
timbastin: Same :)

crowdsourcevexing/crowdsourced_vexing.go
```diff
@@ -0,0 +1,374 @@
...
+				// We only need to check if the paths are not the same
+				if key != crowdsourcedVexRulePath {
+					candidatePathLength, errCandidate := PathPatternMatchLength(inDependencyPath, candidateRule.PathPattern)
+					currentCandidatePathLength, errCurrentCandidate := PathPatternMatchLength(inDependencyPath, currentCandidateRule.PathPattern)
+					if errCandidate != nil || errCurrentCandidate != nil {
+						slog.Error("failed to calculate path pattern match length for tie-breaking", "candidatePath", candidateRule.PathPattern, "currentCandidatePath", currentCandidateRule.PathPattern, "errCandidate", errCandidate, "errCurrentCandidate", errCurrentCandidate)
+						continue
+					}
+					// We check which path covers more dependencies with its wildcard and use that one
+					if candidatePathLength > currentCandidatePathLength {
```
timbastin: I would prefer using the shorter one since its more generic.

Dboy0ZDev: The length counts how many dependencies can be covered by the VEX rule.
I don't quite understand why a shorter length is more generic, shouldn't a path that covers more dependencies with it's wildcard be the more generic one? Am I missing something?

What I think:
```

dependency path: dep1, dep2, dep3, dep4, dep5, dep6
longer path rule: dep2-->*-->dep5 (this has more coverage and should be the more generic one no?)
shorter path rule: dep2-->*-->dep4 (this is a shorter path and has less coverage, so from what I understand less generic)

```

timbastin: This will actually never happen. The path matching is a suffix matching. So only things like: 

* -> A -> B happens. The wildcard has to be the first element of the pattern. 

If we have a dependency tree like this:

A, B, C and A, 1, C there shouldn't be a wildcard like A, *,  C. Using this wildcard we have no idea what functions are really not called or where the vulnerability propagation really ends.

Suffix matching shows the exact location of the inheritance break. *, B, C -> B does not call the vulnerable functions of C. Doesn't matter which package uses B

timbastin: Besides that, we are truth seeking and there has to be ultimately one single truth.

Imagine those two vex rules:

"*", A, B, C and "*", B, C. Those rules are actually telling two whole different stories: the first one tells that A is not calling the vulnerable function BUT B IS vulnerable, while the second rule says B is not vulnerable. If both got voted 50% / 50% it might be a solution to let the program conclude: "actually we still have no idea at all. Those are **contradicting** statements." I am thinking of not using such recommendations at all. If we would present the user the raw challenge, to identify the vulnerability propagation themselves, we might be converging to the real truth in such a uncertain situation

PR #1804, opened 2026-03-23
Refactor vulns and vuln events tables

Hubtrick-Git:
## Vulnerability table changes

- Migrate the id columns of the 3 vuln tables dependency_vulns, first_party_vulnerabilities and license_risk.
- Previously the id was a 256 bit hash stored as a text -> Now its only a 128 bit hash stored as a uuid 
- This leads to 1/4 of the disk space required and performance improvements regarding internal processing of UUIDs in comparison to text

## VulnEvent table changes

- Drop updated_at column to reduce space
- Use 3 foreign key columns, 1 for each vuln instead of the current single column

### Frontend
- [x] Adjust all front end request parameter names
- https://github.com/l3montree-dev/devguard-web/pull/690

### Database
- [x] rebuild indexes

### Queries
- [x] Change all queries using vuln_events
- [x] Change all queries using vuln_ids

### Go Code
- [x] Change vuln event model and refactor affected code
- [x] Change the vulnerability model and refactor affected code

### DIFF
--- a/database/repositories/gorm_repository.go
+++ b/database/repositories/gorm_repository.go
@@ -264,7 +264,7 @@ DELETE FROM vuln_events ve WHERE ve.first_party_vuln_id IS NOT NULL AND NOT EXIS
 	SELECT first_party_vulnerabilities.id FROM first_party_vulnerabilities WHERE first_party_vulnerabilities.id = ve.first_party_vuln_id
 );
 
-DELETE FROM vuln_events ve WHERE ve.license_risk_id = 'licenseRisk' AND NOT EXISTS(
-	SELECT license_risks.id FROM license_risks WHERE license_risks.id = ve.license_risk_id 
+DELETE FROM vuln_events ve WHERE ve.license_risk_id IS NOT NULL AND NOT EXISTS(
+	SELECT license_risks.id FROM license_risks WHERE license_risks.id = ve.license_risk_id
 );
 `


database/migrations/20260323134661_refactor_vuln_ids.up.sql
```diff
@@ -0,0 +1,144 @@
...
+ALTER TABLE artifact_dependency_vulns DROP CONSTRAINT artifact_dependency_vulns_pkey;
+
+-- now we can drop the old dependency vuln id column in the pivot table
+
+ALTER TABLE artifact_dependency_vulns DROP COLUMN dependency_vuln_id;
+
+-- rebuild the primary key of the artifact pivot table with the new column
+
+ALTER TABLE artifact_dependency_vulns ADD PRIMARY KEY (artifact_artifact_name,artifact_asset_version_name,artifact_asset_id,new_dependency_vuln_id);
+
```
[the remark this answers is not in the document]
Hubtrick-Git: will be implemeted

tests/scan_integration_test.go
```diff
@@ -751,6 +754,7 @@ func TestVulnerabilityLifecycleManagement(t *testing.T) {
 			dependencyVulnRepository := f.App.DependencyVulnRepository
 			vulns, _ := dependencyVulnRepository.GetByAssetID(context.Background(), nil, asset.ID)
 			for _, vuln := range vulns {
+				f.DB.Exec("DELETE FROM artifact_dependency_vulns WHERE dependency_vuln_id = ?", vuln.ID)
 				f.DB.Delete(&vuln)
```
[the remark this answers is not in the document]
Hubtrick-Git: test will fail if we encounter an error there, previous delete statements do not check errors either

database/models/vulnevent_model.go
```diff
@@ -22,6 +26,46 @@ type VulnEvent struct {
...
+// SetVulnIDOnEvent sets the appropriate column based on vulnType.
+func SetVulnIDOnEvent(event *VulnEvent, vulnID uuid.UUID, vulnType dtos.VulnType) {
+	switch vulnType {
+	case dtos.VulnTypeDependencyVuln:
+		event.DependencyVulnID = &vulnID
+	case dtos.VulnTypeLicenseRisk:
+		event.LicenseRiskID = &vulnID
+	case dtos.VulnTypeFirstPartyVuln:
+		event.FirstPartyVulnID = &vulnID
+	}
```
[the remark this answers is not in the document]
Hubtrick-Git: This should never be the case and the function only claims to set the vuln id not enforce data integrity to other attributes

database/repositories/vulnevent_repository.go
```diff
@@ -106,9 +106,9 @@ func (r *eventRepository) ReadEventsByAssetIDAndAssetVersionName(ctx context.Con
...
 		Table("vuln_events AS e").
 		Select("e.*, dv.cve_id, dv.component_purl, fv.uri").
-		Joins("LEFT JOIN dependency_vulns dv ON e.vuln_id = dv.id").
-		Joins("LEFT JOIN first_party_vulnerabilities fv ON e.vuln_id = fv.id").
-		Where("(e.vuln_id IN (?) OR e.vuln_id IN (?))", dependencyVulnSubQuery, firstPartyVulnSubQuery).
+		Joins("LEFT JOIN dependency_vulns dv ON e.dependency_vuln_id = dv.id").
+		Joins("LEFT JOIN first_party_vulnerabilities fv ON e.first_party_vuln_id = fv.id").
+		Where("(e.dependency_vuln_id IN (?) OR e.first_party_vuln_id IN (?))", dependencyVulnSubQuery, firstPartyVulnSubQuery).
 		Order("e.created_at DESC").
 		Find(&events)
```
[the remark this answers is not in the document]
Hubtrick-Git: this logic was implemented independently of my refactor and so i will not touch it .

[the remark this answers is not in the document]
Hubtrick-Git: It did not list them previously so I do not think there is a point in adding that now

database/repositories/vulnevent_repository.go
```diff
@@ -24,17 +24,17 @@ func NewVulnEventRepository(db *gorm.DB) *eventRepository {
 	}
 }
 
-func (r *eventRepository) ReadAssetEventsByVulnID(ctx context.Context, tx *gorm.DB, vulnID string, vulnType dtos.VulnType) ([]models.VulnEventDetail, error) {
+func (r *eventRepository) ReadAssetEventsByVulnID(ctx context.Context, tx *gorm.DB, vulnID uuid.UUID, vulnType dtos.VulnType) ([]models.VulnEventDetail, error) {
 	if vulnType == dtos.VulnTypeDependencyVuln {
 		return r.readDependencyVulnAssetEvents(ctx, tx, vulnID)
 	}
 	return r.readFirstPartyVulnAssetEvents(ctx, tx, vulnID)
 }
```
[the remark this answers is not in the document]
Hubtrick-Git: The previous logic did not include license risk so I will not add such a breaking change to a refactor which is independent of this.

database/migrations/tooling-read-me.txt
timbastin, line 1: How about we put this in the regular readme?

database/migrations/20260323134662_vuln_events_disk_space_and_relations_improvement.down.sql
timbastin, line 1: Can we combine those two migrations into one? Otherwise the database is in an invalid state between the two migration runs.

database/migrations/20260323134661_refactor_vuln_ids.up.sql
```diff
@@ -0,0 +1,148 @@
...
+ALTER TABLE public.artifact_license_risks ADD FOREIGN KEY (new_license_risk_id) REFERENCES public.license_risks (new_id);
+
+-- Finally rename columns in clean up process
+ALTER TABLE public.license_risks RENAME COLUMN new_id TO id;
+ALTER TABLE public.artifact_license_risks RENAME COLUMN new_license_risk_id TO license_risk_id;
+
+
+-- Create indexes at the end
+DROP INDEX IF EXISTS public.idx_artifact_dependency_vulns_dependency_vuln_id;
+CREATE INDEX idx_artifact_dependency_vulns_dependency_vuln_id ON public.artifact_dependency_vulns USING btree (dependency_vuln_id);
```
timbastin: Nitpicking, but a hash index should be faster right here. Since its a uuid, we will only do = queries - no ranges

database/models/vulnevent_model.go
```diff
@@ -22,6 +26,46 @@ type VulnEvent struct {
...
+		return dtos.VulnTypeLicenseRisk
+	}
+	if event.FirstPartyVulnID != nil {
+		return dtos.VulnTypeFirstPartyVuln
+	}
+	return ""
+}
+
+// SetVulnIDOnEvent sets the appropriate column based on vulnType.
+func SetVulnIDOnEvent(event *VulnEvent, vulnID uuid.UUID, vulnType dtos.VulnType) {
```
timbastin, line 58: this right here should be done in the transformer package. We will get issues when importing dtos into models.

timbastin: Maybe its simpler to move the vulntype to the models package itself.

database/repositories/artifact_repository.go
```diff
@@ -68,3 +77,45 @@ func (r *artifactRepository) GetAllArtifactAffectedByDependencyVuln(ctx context.
 	}
 	return artifacts, nil
 }
+
+var CleanupOrphanedRecordsSQL = `
```
timbastin: I think i moved this into its own daemon

database/repositories/gorm_repository.go
```diff
@@ -226,45 +226,3 @@ func isIgnorableUpsertError(err error) bool {
 
 	return false
 }
-
-var CleanupOrphanedRecordsSQL = `
```
timbastin, line 230: Okay - not a daemon but gorm_repository. What was the reason to move this back to artifact_repository. This handles plenty of different tables therefore i was moving it to a more generic file.

timbastin (changes requested):

database/migrations/tooling-read-me.txt
Hubtrick-Git: I think it is a very specific command only useful for developers. Having it in the migrations folder directly helps those who are confused on how to create migrations and directly helps them in this case.

database/migrations/20260323134662_vuln_events_disk_space_and_relations_improvement.down.sql
Hubtrick-Git: Yes thats a very good point

database/migrations/20260323134661_refactor_vuln_ids.up.sql
Hubtrick-Git: Yes I still used a btree for concurrent hash creation which would not lock the database during its creation. But since we do not want that, hash indexes are infact a tiny bit faster and require less space.

database/repositories/gorm_repository.go
Hubtrick-Git: When I merged I saw both options and the option in the artifact repository used a updated logic -> I interpreted it as the most recent implementation and therefore the right one. 

Will revert this change to the state on main.

database/models/vulnevent_model.go
Hubtrick-Git: I dont really understand the 2nd comment. It is currently in the models package. Besides that we regularly import the dtos package in the models package.

PR #1874, opened 2026-04-12
fix(1872): added fix for vex rule not working in case of direct dependancy

gauravshinde1729:
# Title: VEX rules not applied for direct dependencies (https://github.com/l3montree-dev/devguard/issues/1872)

  # Summary:

  Fixed a bug in matchPatternExact where patterns like ["*", "ROOT", "pkg:..."] never matched direct dependency vulnerability paths
  like ["pkg:..."].

  When a wildcard (*) was processing and its next pattern element was also a wildcard (ROOT), the code searched for "ROOT" literally
  in the path. Since ROOT is a structural graph node and never stored in VulnerabilityPath, the match always failed.

  Fix: Before searching for the next pattern element literally, check if it is itself a wildcard. If so, attempt a zero-element match
   by recursing directly — consistent with how IsWildcard already treats ROOT.

  Files changed:
  - `dtos/vex_rule_dto.go` — fixed matchPatternExact to handle consecutive wildcards
  - `dtos/path_pattern_test.go` — added test case for direct dependency matching

Output:

Screenshot 2026-04-12 075254

Screenshot 2026-04-12 080003

Screenshot 2026-04-12 080307

### DIFF
--- a/dtos/path_pattern_test.go
+++ b/dtos/path_pattern_test.go
@@ -100,6 +100,9 @@ func TestRootPathPattern(t *testing.T) {
 		{"ROOT matches any path with ROOT at end", PathPattern{normalize.GraphRootNodeID}, []string{"A", "B", normalize.GraphRootNodeID}, true},
 		{"ROOT DOES match path without ROOT", PathPattern{normalize.GraphRootNodeID}, []string{"A", "B", "C"}, true},
 		{"ROOT does not lead to all paths matching", PathPattern{normalize.GraphRootNodeID, "X"}, []string{"A", "B", "C"}, false},
+		// Direct dependency: pattern created from the graph includes ROOT but VulnerabilityPath does not.
+		// ["*", "ROOT", "pkg:..."] must match ["pkg:..."] because ROOT is a wildcard that matches zero elements.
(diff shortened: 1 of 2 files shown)


timbastin:
Hi @gauravshinde1729 thanks a lot for your PR! I think there is just a logical issue right here. Root actually is not a wildcard. It is used as a stop marker, so the vulnerability path needs to end actually. This means, that [root, pkg:A] should match the path [pkg:A].

The pattern:
[*, root, pkg:A] is equivalent to the pattern [root, pkg:A].

[root, pkg:A] **should not** match [pkg:B, pkg:A]. This is different to the star symbol. [*, pkg:A] **should** match [pkg:B, pkg:A].

Would it be possible for you to make the changes accordingly?

gauravshinde1729:
Hey @timbastin 
Sure , I will recheck the logic once 
Thanks !

gauravshinde1729:
Hey @timbastin 

I have added fix for this and test changes for both transitive & direct dependency 



image
image

Can you please check once

Thanks !

timbastin (approved):

timbastin:
Thanks a lot! This is great!

PR #1882, opened 2026-04-15
Add dependency proxy configs

refoo0:


controllers/dependency_proxy_controller.go
```diff
@@ -44,23 +46,50 @@ const (
 	PyPIProxy ProxyType = "pypi"
 )
 
-type DependencyProxyConfig struct {
+type DependencyProxyCache struct {
 	CacheDir string
 }
+type DependencyProxyConfigs struct {
+	Rules          []string `json:"rules"`
+	MinReleaseTime int      `json:"minReleaseTime"` // in hours
```
timbastin:
```suggestion
	MinReleaseAge int      `json:"minReleaseAge"` // in hours
```

cmd/devguard-cli/commands/vulndb.go
```diff
@@ -88,6 +88,7 @@ func migrateDB() {
 			fx.Invoke(func(ShareRouter router.ShareRouter) {}),
 			fx.Invoke(func(VulnDBRouter router.VulnDBRouter) {}),
 			fx.Invoke(func(dependencyProxyRouter router.DependencyProxyRouter) {}),
+			fx.Invoke(func(shareDependencyProxyRouter router.ShareDependencyProxyRouter) {}),
```
timbastin: I dislike the name, since it has nothing todo with really sharing something. Can't we just put the routes inside the DependencyProxyRouter itself?

controllers/dependency_proxy_controller.go
```diff
@@ -44,23 +46,50 @@ const (
...
+}
 
 type DependencyProxyController struct {
-	maliciousChecker shared.MaliciousPackageChecker
-	cacheDir         string
-	client           *http.Client
+	assetRepository        shared.AssetRepository
+	projectRepository      shared.ProjectRepository
+	orgRepository          shared.OrganizationRepository
+	dependencyProxyService shared.DependencyProxySecretService
```
timbastin: 😎

controllers/dependency_proxy_controller.go
```diff
@@ -44,23 +46,50 @@ const (
...
+	maliciousChecker       shared.MaliciousPackageChecker
+	cacheDir               string
+	client                 *http.Client
+}
+
+// trimProxyPrefix strips the /api/v1/dependency-proxy/[secret/]<ecosystem> prefix from the path.
+// The secret segment is optional to support routes with and without a secret.
+func TrimProxyPrefix(path string, ecosystem ProxyType) string {
+	encodedPackage := regexp.MustCompile(`^/api/v1/dependency-proxy/(?:[^/]+/)?`+regexp.QuoteMeta(string(ecosystem))+`(?:/|$)`).ReplaceAllString(path, "")
```
timbastin: Lets move the MustCompile outside the function call itself. I am afraid, that we did not test an ecosystem and have a runtime panic. I think we can just hard code the 3 different regexes.

controllers/dependency_proxy_controller.go
```diff
@@ -89,41 +126,68 @@ func (d *DependencyProxyController) ProxyNPM(c shared.Context) error {
...
 
-	if blocked, reason := d.checkMaliciousPackage(ctx, NPMProxy, requestPath); blocked {
-		slog.Warn("Blocked malicious package", "proxy", "npm", "path", requestPath, "reason", reason)
-		// Also remove from cache if it exists to prevent serving cached malicious content
-		cachePath := d.getCachePath(NPMProxy, requestPath)
-		if err := os.Remove(cachePath); err == nil {
-			slog.Info("Removed malicious package from cache", "path", cachePath)
+	if hasExplicitVersion {
+
+		notAllowed, notAllowedReason := d.CheckNotAllowedPackage(ctx, NPMProxy, requestPath, configs)
```
timbastin: We should add a log statement right here - same like in the malicious packages branch

controllers/dependency_proxy_controller.go
```diff
@@ -148,12 +212,25 @@ func (d *DependencyProxyController) ProxyNPM(c shared.Context) error {
...
+	if !hasExplicitVersion {
+		//check allowlist patterns before checking malicious database to prevent false positives on allowed packages
+		notAllowed, notAllowedReason := d.CheckNotAllowedPackage(ctx, NPMProxy, packageName+"@"+resolvedVersion, configs)
+		if notAllowed {
+			return d.blockNotAllowedPackage(c, NPMProxy, requestPath, notAllowedReason)
+		}
+
 		// Parse the JSON response to extract the version that would be installed
-		if resolvedVersion := d.ExtractNPMVersionFromMetadata(data); resolvedVersion != "" {
+		if resolvedVersion != "" && d.maliciousChecker != nil {
```
timbastin: I think we should fail - even panic if d.maliciousChecker IS nil. It should NEVER be nil. This if statement will result in: "There is a bug in devguard itself, this will not THROW and silently disable the dependency proxy firewall without telling the user"

router/share_dependency_proxy_router.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func NewShareDependencyProxyRouter(
+	apiV1Group APIV1Router,
+	dependencyProxyController *controllers.DependencyProxyController,
+) ShareDependencyProxyRouter {
+	shareDependencyProxyRouter := apiV1Group.Group.Group("/dependency-proxy/:secret")
+
+	shareDependencyProxyRouter.GET("/npm", dependencyProxyController.ProxyNPM)
+	shareDependencyProxyRouter.GET("/npm/*", dependencyProxyController.ProxyNPM)
```
timbastin: Lets make this explicit. I think it is super hard to follow the code, since each proxy function handles different requests. One for tar.gz, metadata requests etc. I think it would be much simpler to understand this security critical part of devguard after splitting each route into its own dedicated function. What do you think?

timbastin (changes requested):

PR #1890, opened 2026-04-21
Rework vulndb import

Hubtrick-Git:
description coming soon...


tests/scan_integration_test.go
Hubtrick-Git, line 1: Here the tests still fail due to 2 functions not getting called properly. When trying to debug this problem I  ran into some non deterministic panics and go routine debugger problems, which prevented me from further handling this test failure.

timbastin:
ecosystems.zip has a size of 180MB. An uncompressed size of 1.3GB

Using gob I was able to reduce the size to around 450MB. Compressing that with the zst algorithm leads to 71MB of size. I think that is all we can do...

PR #1893, opened 2026-04-22
updates Readme

l3monKenji:
Please check if the Readme update is okay for you.

It adds the new Features and includes a Quick-Start (please check also if it's technically correct, since i took it from the documentation and reworked it a bit).

Also the new Partner Logos are added.


timbastin (approved):

package-lock.json
timbastin, line 1: I think this can be deleted

timbastin (changes requested):

PR #1907, opened 2026-04-27
adds user agent check for mcp-server in CreateEvent method

refoo0:


database/migrations/20260427151344_add_user_agent_column_to_events_table.down.sql
timbastin, line 1: Since we are using PostgreSQL we don't need the down migrations at all. We just removed all down migrations from the main branch.

integrations/commonint/integration_helper.go
```diff
@@ -145,13 +145,13 @@ func CreateNewVulnEventBasedOnComment(vulnID uuid.UUID, vulnType dtos.VulnType,
 
 	switch event {
 	case dtos.EventTypeAccepted:
-		return models.NewAcceptedEvent(vulnID, vulnType, userID, justification, false)
+		return models.NewAcceptedEvent(vulnID, vulnType, userID, justification, false, "")
```
timbastin: I think we should be consistent about what the null value of the user agent should be. There is a check in this PR which checks if the user agent is nil, besides that, the migration does not provide any return value. Let's stick to nil as the null value for the user agent not "". What do you think?

integrations/gitlabint/gitlab_webhook.go
timbastin, line 1: We can use the webhook request user agent right here

integrations/jiraint/jira_integration.go
timbastin, line 1: Same here, let's use the real user agent

timbastin (changes requested):

timbastin (approved):

PR #1912, opened 2026-04-29
Implemented crowdsourced vexing

Dboy0ZDev:
Integrated crowdsourced vexing and added endpoint
Issue: https://github.com/l3montree-dev/devguard/issues/1903


database/models/trusted_entities_model.go
```diff
@@ -7,12 +7,12 @@ import (
 )
 
 type TrustedEntity struct {
-	ID             uuid.UUID  `json:"id" gorm:"type:uuid;primaryKey;default:gen_random_uuid()"`
-	OrganizationID *uuid.UUID `json:"organization_id" gorm:"type:uuid;default:null;uniqueIndex:idx_trusted_entities_unique"`
-	ProjectID      *uuid.UUID `json:"project_id" gorm:"type:uuid;default:null;uniqueIndex:idx_trusted_entities_unique"`
-	TrustScore     float64    `json:"trustScore" gorm:"not null;default:0"`
-	CreatedAt      time.Time  `json:"createdAt" gorm:"autoCreateTime"`
-	UpdatedAt      time.Time  `json:"updatedAt" gorm:"autoUpdateTime"`
+	TrustedEntityID uuid.UUID  `json:"trusted_entity_id" gorm:"type:uuid;primaryKey;default:gen_random_uuid()"`
```
Hubtrick-Git: Maybe you could replace the id primary key with a composite key on organization ID and project ID. This saves 1 column and 1 index.

database/repositories/trusted_entities_repository.go
Hubtrick-Git, line 28: We could use "on conflict id update set ..." here instead of manually handling this case.

Issue #1921, opened 2026-05-04
OCI Proxy Sanitization


seb-kw:
The /v2/... OCI proxy is unauthenticated and accepts any upstream host. When that host returns 401 + WWW-Authenticate: Bearer realm="<URL>", devguard fetches <URL> with no validation — letting an unauthenticated remote attacker drive devguard's outbound HTTP client at internal targets (cloud metadata, RFC1918, loopback).

#### Reproduce

```
# Unauth reachable
curl -i http://<devguard>/v2/                                          # 200 {}

# Attacker host accepted; example.com's response is proxied back
curl -i http://<devguard>/v2/example.com/foo/manifests/latest          # 404 from cloudflare

# Hosting `evil.com` returning  WWW-Authenticate: Bearer realm="http://169.254.169.254/..."
# and calling /v2/evil.com/foo/manifests/latest causes devguard to GET 169.254.169.254
```

#### Root cause
`controllers/dependencyfirewall/oci.go:199-237` — fetchRegistryToken uses realm verbatim.
`router/oci_registry_router.go:50-74` — /v2/... group has no auth middleware.
`utils/egress_client.go:26-45` — Egress Transport does not block private/loopback/link-local IPs.

#### Proposed Improvements

- [ ] Add DEPENDENCY_PROXY_OCI_PUBLIC_ENABLED env (default true); when false, do not register the unauth /v2/:registry/... routes.
- [ ] Add DEPENDENCY_PROXY_OCI_REGISTRIES allowlist; reject :registry values that are IP literals, contain :, or aren't on the list (400) - empty is allow all.
- [ ] Allowlist Content-Type per route (manifests / blobs / tags-list); reject mismatch with 502.
- [ ] Require Docker-Distribution-API-Version: registry/2.0 on every upstream response.
- [ ] Validate route params with OCI regexes: :image/:namespace (name), :reference (tag-or-digest), :digest (^sha256:[a-f0-9]{64}$).
- [ ] CheckRedirect on the OCI client: allow only original upstream host or explicit CDN allowlist.
- [ ] In fetchRegistryToken, validate realm: scheme https, host == upstream host or in a registry→auth-host map.
- [ ] Augment EgressTransport with a DialContext that rejects RFC1918, 127.0.0.0/8, 169.254.0.0/16, ::1, fc00::/7, fe80::/10.

####  References

OCI Distribution Spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md

```
[...]

PR #1928, opened 2026-05-06
Harden OCI proxy against SSRF (#1921)

gauravshinde1729:
https://github.com/l3montree-dev/devguard/issues/1921

Changes made in the PR:

  1. Check the realm before fetching a token

  A bad registry could return a fake realm URL pointing to a private IP (like AWS metadata). The proxy used to follow it blindly. Now we check the realm first:

  - it must be a valid URL
  - it must use HTTPS
  - it must point to the same host as the registry, or a host we trust for that registry

  If any check fails, the token call is never made.

  2. Allowlist of registries:

The :registry part of the URL used to accept anything. Now only seven public registries are allowed: docker.io, ghcr.io, quay.io, gcr.io, registry.k8s.io, public.ecr.aws, mcr.microsoft.com. IP addresses and host:port forms are always blocked.

  Test: /v2/evil.com/... → 400.     /v2/1.2.3.4/... → 400.

  3. One file for all registry config:

  Registry info used to live in three places. Combined them into oci_registries.go — one list, three maps built from it. Adding a new registry is one line.

  Test: A small test makes sure every entry shows up in every map.

  4. Fix the trailing-slash bug:

  While testing, we found a global middleware adding / to every URL. This broke OCI calls because the spec doesn't allow trailing slashes. The middleware now skips /v2/ paths.

  Test: All seven registries return success end-to-end through the proxy.

  5. Validate path parameters:

  Image names, tags, and digests must now match the OCI spec format. Catches path traversal or short digests early.

  Test: ../../../etc/passwd in the URL → 400. Short digest → 400.

  6. Kill switch for the public proxy:

 Set DEPENDENCY_PROXY_OCI_PUBLIC_ENABLED=false to disable the public route entirely. Default leaves it on. Secret-scoped routes are unaffected.

 Test: With the env var set, public URL returns 404. Secret-scoped URLs still work.

  ---
  Order of checks at runtime

  1. Is the route registered? 
  2. Is the registry allowed? (
  3. Are the path params valid? 
  4. Cache hit? If not, call the upstream.
  5. If upstream returns 401, is the realm safe ?
  6. Get token, retry, return.

  How everything fits together at runtime

  Request → /v2/<registry>/<image-path>/<manifests|blobs|...>
              │
              ▼
      Route registered?              
              │
              ▼
      isAllowedRegistry(registry)    
              │
              ▼
      validateOCIPathParams(c)
	      │
              ▼
      Check cache / upstream call
              │
              ▼ (on 401)
      validateRealmHost(realm,host)  ← Gate 6 (the SSRF fix)
              │
              ▼
      fetchToken → retry → return

  Each layer fails fast with a clear 400 or 502 and never reaches the next.

### DIFF
--- a/controllers/dependencyfirewall/oci.go
+++ b/controllers/dependencyfirewall/oci.go
@@ -40,6 +40,18 @@ import (
 
 var ociProxyPrefixRe = regexp.MustCompile(`^/api/v1/dependency-proxy/(?:[^/]+/)?oci(?:/|$)`)
 
+// OCI Distribution Spec path-param formats. Anything outside these is
+// rejected at the edge so that path traversal, NUL injection, or other
+// surprises never reach the cache layer or the upstream registry.
+var (
+	// One image-name segment (e.g. "library", "nginx").
+	ociNameSegmentRe = regexp.MustCompile(`^[a-z0-9]+(?:(?:[._]|__|[-]+)[a-z0-9]+)*$`)
+	// Tag: 1-128 chars, alphanumeric/_/./-, must not start with . or -.
+	ociTagRe = regexp.MustCompile(`^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$`)
+	// Digest: only sha256 supported by the proxy today.
+	ociDigestRe = regexp.MustCompile(`^sha256:[a-f0-9]{64}$`)
+)
+
 // OCIDependencyProxyController handles OCI registry proxy requests.
 // Image references must be fully qualified: <registry>/<image> (e.g. docker.io/library/nginx).
 // It embeds DependencyProxyController to reuse shared helpers and state.
@@ -145,6 +157,32 @@ func isAllowedRegistry(registry string) bool {
 		return false
 	}
 	return slices.Contains(allowedOCIRegistries, registry)
+}
+
+// validateOCIPathParams enforces the OCI Distribution Spec format on every
+// path-param Echo gives us. Anything malformed (path traversal, special
+// characters, wrong digest length) is refused before the proxy touches the
+// cache or the upstream registry.
+func validateOCIPathParams(c shared.Context) error {
+	for _, segment := range []string{c.Param("image"), c.Param("namespace"), c.Param("ns1"), c.Param("ns2")} {
+		if segment == "" {
+			continue
+		}
+		if !ociNameSegmentRe.MatchString(segment) {
+			return echo.NewHTTPError(http.StatusBadRequest, "invalid image name")
+		}
+	}
+	if ref := c.Param("reference"); ref != "" {
+		if !ociTagRe.MatchString(ref) && !ociDigestRe.MatchString(ref) {
+			return echo.NewHTTPError(http.StatusBadRequest, "invalid reference")
+		}
+	}
+	if dig := c.Param("digest"); dig != "" {
+		if !ociDigestRe.MatchString(dig) {
+			return echo.NewHTTPError(http.StatusBadRequest, "invalid digest")
+		}
+	}
+	return nil
 }
 
 // upstreamURLForRegistry returns the upstream base URL for a given registry
@@ -373,6 +411,9 @@ func (d *OCIDependencyProxyController) ProxyOCIManifest(c shared.Context) error 
 	if !isAllowedRegistry(registry) {
 		return echo.NewHTTPError(http.StatusBadRequest, "registry not allowed")
 	}
+	if err := validateOCIPathParams(c); err != nil {
+		return err
+	}
 	reference := c.Param("reference")
 	// Path sent to the upstream (no registry prefix).
 	upstreamPath := fmt.Sprintf("/v2/%s/manifests/%s", upstreamImagePath, reference)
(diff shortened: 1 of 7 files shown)


controllers/dependencyfirewall/oci.go
```diff
@@ -130,11 +144,53 @@ func ociSafeCachePath(requestPath string) string {
...
 
-// upstreamURLForRegistry returns the upstream base URL for a given registry hostname.
-// docker.io is a special case: its actual API endpoint differs from the pull hostname.
+// isAllowedRegistry reports whether requests for the given :registry path
+// segment may proceed. IP literals and host:port forms are rejected outright
+// — they are never legitimate public-registry hostnames and have historically
+// been the SSRF entry point. Otherwise the registry must be one of the
+// supported registries listed in oci_registries.go.
+func isAllowedRegistry(registry string) bool {
+	if net.ParseIP(registry) != nil {
```
timbastin: You can skip this if statement and the one on line 156. The last condition on line 159 is already enough. Since the allowedOCIRegistries do not contain any IP nor a ":" that check is sufficient.

controllers/dependencyfirewall/oci.go
```diff
@@ -130,11 +144,53 @@ func ociSafeCachePath(requestPath string) string {
...
+		return false
+	}
+	return slices.Contains(allowedOCIRegistries, registry)
+}
+
+// validateOCIPathParams enforces the OCI Distribution Spec format on every
+// path-param Echo gives us. Anything malformed (path traversal, special
+// characters, wrong digest length) is refused before the proxy touches the
+// cache or the upstream registry.
+func validateOCIPathParams(c shared.Context) error {
```
timbastin: What do you think about reusing the already existing imageParamsFromContext right here? Maybe even just accepting registry, fqImageName, and upstreamImagePath as parameters so that we only extract them once, since it is already rather difficult to grasp what n1 and n2 are.

controllers/dependencyfirewall/oci_registries.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Lookup tables derived from supportedOCIRegistries in init().
+var (
+	allowedOCIRegistries  []string            // pathNames accepted at the proxy edge
+	registryAuthHosts     map[string][]string // upstreamHost → permitted token-issuer hosts
+	registryUpstreamHosts map[string]string   // pathName → upstreamHost
+)
+
+func init() {
```
timbastin, line 55: I think this can be simplified by just inlining the maps directly without creating the dedicated supportedOCIRegistry struct at all

timbastin: would be great to have registry.gitlab.com and registry.opencode.de here as well.

middlewares/server.go
```diff
@@ -40,7 +41,15 @@ func registerMiddlewares(e *echo.Echo) {
 	// Expose the trace ID to the client so it can be referenced in Jaeger / GlitchTip.
 	e.Use(traceID())
 
-	e.Pre(middleware.AddTrailingSlash())
+	// AddTrailingSlash normalises REST endpoints, but it must be skipped for
+	// the OCI Distribution Spec routes — /v2/<name>/manifests/<reference> and
+	// friends are defined without trailing slashes, and adding one causes
+	// every registry (ghcr.io, quay.io, ...) to return 404.
+	e.Pre(middleware.AddTrailingSlashWithConfig(middleware.TrailingSlashConfig{
+		Skipper: func(c echo.Context) bool {
```
timbastin, line 49: GREAT!!

timbastin (changes requested):
Awesome work! Thanks a lot for this PR. I just have some minor request for changes. Love it ❤️

middlewares/server.go
gauravshinde1729: Thanks !

controllers/dependencyfirewall/oci_registries.go
gauravshinde1729: Hey @timbastin  I agree on simplifying the implementation. However, I’m keen on keeping this in a single registry file to ensure centralized management. This makes it much easier to track and extend later on. I’ll go ahead and swap the struct for a simple map as you suggested, but I’d like to keep it in its own file for better maintainability. What's your take on this ! 


Also for registry.gitlab.com realm host is ` https://gitlab.com/jwt/auth.`  but for  registry.opencode.de, do we know the realm host ? so that I can keep it in our oci registry ?

Thanks !

controllers/dependencyfirewall/oci.go
gauravshinde1729: Hey  @timbastin , agreed ! , I handled this in latest commit 

Thanks !

controllers/dependencyfirewall/oci.go
gauravshinde1729: Sure, I agree with this point , I have made the changes in the latest commit 

Thanks !

gauravshinde1729:
Hey @timbastin  , I have addressed the comments , can you please check once 
Thanks !

timbastin:
Looks great! Thanks a lot!

timbastin (approved):

PR #1936, opened 2026-05-06
add config option to disable organization creation

### LINKED ISSUE #1720: Add config option to disable organisation creation for an DevGuard instance
Would be nice to have a env or something (configurable via Helm Chart) to disable further org creation for scenarios where admins want to only provide a single org on a instance.

refoo0:


controllers/org_controller.go
```diff
@@ -56,6 +58,15 @@ func NewOrganizationController(repository shared.OrganizationRepository, orgServ
 // @Success 200 {object} models.Org
 // @Router /organizations [post]
 func (controller *OrgController) Create(ctx shared.Context) error {
+	var settings shared.InstanceSettings
+	err := controller.configService.GetJSONConfig(ctx.Request().Context(), "instance_settings", &settings)
```
timbastin: We should use camelCase right here. I would prefer instanceSettings.

Maybe we can even handle this inside a middleware. This way settings could just disable endpoints. This would be a cleaner architecture in my opinion

daemons/daemon.go
timbastin, line 1: I think it makes no sense to modify settings dynamically on each daemon run. We can have something like: all env variables will get loaded on startup and will always populate the instance settings. But this makes sense as a startup task not iteratively

timbastin (changes requested):

timbastin (approved):

PR #1950, opened 2026-05-10
Feat/implement cache stats endpoint

gauravshinde1729:
https://github.com/l3montree-dev/devguard/issues/1915

Adds a new admin endpoint to inspect on-disk consumption of the dependency and OCI proxy cache.

`GET /api/v1/organizations/:organization/dependency-proxy/cache-stats/`

Gated to org admins (`manage` scope + `ActionUpdate` on the organization), matching the existing org-stats endpoint. Walks the configured cache directory once and returns aggregate storage usage — overall, per ecosystem (`npm`, `go`, `pypi`, `oci`), and an OCI manifest-vs-blob split.

Companion files written by the proxy (`.sha256`, `.releasetime`,   `.contenttype`, `.digest`) count toward `sizeBytes` but **not** toward `entries`, so the entry count reflects real cached payloads only. Ecosystems that have never been hit don't appear in `byEcosystem`. A missing cache directory returns zeroed stats rather than an error.

  Example response:

  ```json
  {
    "cacheDir": "/tmp/devguard-dependency-proxy-cache",
    "totalSizeBytes": 825215,
    "totalEntries": 4,
    "oldestEntry": "2026-05-10T10:02:44Z",
    "newestEntry": "2026-05-10T22:19:52Z",
    "byEcosystem": {
      "npm": {
        "sizeBytes": 801366,
        "entries": 1,
        "oldestEntry": "2026-05-10T22:09:26Z",
        "newestEntry": "2026-05-10T22:09:26Z"
      },
      "oci": {
        "sizeBytes": 23849,
        "entries": 3,
        "oldestEntry": "2026-05-10T10:02:44Z",
        "newestEntry": "2026-05-10T22:19:52Z"
      }
    },
    "ociBreakdown": {
      "manifests": { "sizeBytes": 23849, "entries": 3 },
      "blobs":     { "sizeBytes": 0,     "entries": 0 }
    }
  }

### DIFF
--- a/controllers/dependencyfirewall/cache_stats.go
+++ b/controllers/dependencyfirewall/cache_stats.go
@@ -1,0 +1,189 @@
+// Copyright (C) 2026 l3montree GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package dependencyfirewall
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/l3montree-dev/devguard/shared"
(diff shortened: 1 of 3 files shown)


timbastin:
I really like this! The only concern I have is, that currently we don't have a separate cache per Organisation. There is just a single global dependency proxy cache. We should either:

- make the cache org scoped, so that we can iterate over a subdirectory 
- wait until the instance admin feature is fully implemented (you can find an open pr for this) and let only instance admin see the total cache for all created organizations

In the end, it would be great to have both. So for you to start straight away maybe you can work on creating separate sub folders for the cache directory for each Organisation. We have to watch out for cache poisoning. Maybe we can use the org id as folder name. What do you think?

gauravshinde1729:
Sure Thanks @timbastin , That's a great idea !

timbastin:
@gauravshinde1729 I will close this due to inactivity. If you are keen to continue or work, just let me know!

PR #2035, opened 2026-05-20
Ingest upstream vex sources

Dboy0ZDev:
Related to Issue: https://github.com/l3montree-dev/devguard/issues/1977


services/vex_rule_service.go
```diff
@@ -492,18 +493,14 @@ func (s *VEXRuleService) parseVEXRulesInBOM(ctx context.Context, assetID uuid.UU
...
-			// If no metadata component PURL, use the affected package directly
-			pattern = dtos.PathPattern{purlString}
+		componentPurlStr, err := normalize.PURLToString(componentPurl)
+		if err != nil {
+			slog.Info("failed to unescape component purl for path pattern, continuing anyway", "purl", componentPurl.String(), "error", err)
+			componentPurlStr = componentPurl.String()
 		}
 
+		pattern = dtos.PathPattern{componentPurlStr, dtos.PathPatternWildcard, purlString}
+
```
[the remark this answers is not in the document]
Dboy0ZDev: 1. The removed check only guards for empty `componentPurl` and not invalid purls anyways
2. This check is already guarded through a check above, where the `componentPurl` var is extracted
```
    var componentPurl packageurl.PackageURL
	if bom.Metadata.Component.PackageURL == "" {
		return nil, fmt.Errorf("no package URL found in metadata component")
	}

	var err error
	componentPurl, err = packageurl.FromString(bom.Metadata.Component.PackageURL)
	if err != nil {
		slog.Info("failed to parse metadata component PURL, continuing anyway", "purl", bom.Metadata.Component.PackageURL, "error", err)
	}
```
The componentPurl cannot be empty at this point.

services/scan_service.go
```diff
@@ -864,3 +877,120 @@ func (s *scanService) ScanSBOMWithoutSaving(ctx context.Context, bom *cyclonedx.
...
+
+func (s *scanService) FetchOpenVexFromGitHub(ctx context.Context, targetURL string) (vexReports []*normalize.VexReportOpenVEX, err error) {
+	client := newGitHubClient()
+	owner, repo, err := ParseGitHubURL(targetURL)
+	if err != nil {
+		return nil, err
+	}
+
+	// Determine default branch
+	repository, _, err := client.Repositories.Get(ctx, owner, repo)
```
timbastin: I know, that github has rather strict rate limits. What do you think about downloading the whole repo once, maybe via zip, and then parse it.

normalize/sbom_graph.go
```diff
@@ -141,6 +142,41 @@ func NewVexReport(report *cdx.BOM, source string) (*VexReport, error) {
...
+	if report.Timestamp == nil || report.Timestamp.IsZero() {
+		return fmt.Errorf("invalid OpenVEX report: missing timestamp")
+	}
+	if report.Version == 0 {
+		return fmt.Errorf("invalid OpenVEX report: missing version")
+	}
+	return nil
+}
+
+func NewVexReportOpenVEX(report *ov.VEX, source string) (*VexReportOpenVEX, error) {
```
timbastin: Do you think its possible to convert openVEX to cyclonedx? This way we could just reuse all the already existing functions. Or we transform cyclonedx to openvex (which might be easier since cyclonedx has a larger specification than openvex).

timbastin (changes requested):

services/scan_service.go
Dboy0ZDev: Didn't know this was an option, but now that you mention it, makes sense. Will implement it instead of the download loop 👍 

Should the zip then also be used to determine the target files and the branch or can those request be kept as they are? 

```
repository, _, err := client.Repositories.Get(ctx, owner, repo)

//...

tree, _, err := client.Git.GetTree(
		ctx,
		owner,
		repo,
		branch,
		true, // recursive
	)
```

I think those can/should remain as they are. This would leave us as 3 requests total for each repo scan instead of the 2 + number_of_vex_documents.

normalize/sbom_graph.go
Dboy0ZDev: Possible, but since openvex documents are not as granular as cyclonedx documents, the resulting converted cyclonedx document could have missing/empty fields. 

E.g. OpenVex providesimpact statements `affected`,`not_affected`,`fixed` and `under_investigation`, while Cyclonedx provides more granularity with analysis states of `false_positive`,`not_affected`,`exploitable`, `resolved`,`resolved_with_pedigree`, `in_triage`.

Another issue could be that the current implementation of the cyclonedx parsing uses BOM information that is included in a cyclonedx document, but openvex does not provide BOM information, just pure Vexing results.

timbastin: okay then lets keep it as is to have a clear split instead of many unknown defaults

services/scan_service.go
timbastin: Actually I think its perfectly fine to expect the branch name as a parameter of this function. I am pretty sure you can walk over the zip file in memory. Checkout utils/zip.go - there you get a reader of a whole zip file without even saving it on the disk. Try to keep it in memory. Then you can walk over that and you dont need the git file tree at all

Dboy0ZDev: Changes have been implemented, have a look.

daemons/openvex_daemon.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"log/slog"
+	"os"
+	"strings"
+
+	"github.com/l3montree-dev/devguard/normalize"
+)
+
+func (runner *DaemonRunner) UpdateSystemVEXRulesFromOpenVEXSources(ctx context.Context) error {
+	enviromentSources := os.Getenv("OPENVEX_SOURCES")
```
Hubtrick-Git: Your pull request does not modify the .env.example, so I do not know of this env or what to use as a value.

database/migrations/20260728190001_add_upstream_vex_rules.up.sql
```diff
@@ -0,0 +1,36 @@
...
+CREATE TABLE IF NOT EXISTS public.system_vex_rules (
+    id TEXT PRIMARY KEY,
+    cve_id TEXT NOT NULL,
+    justification TEXT NOT NULL,
+    mechanical_justification TEXT,
+    path_pattern JSONB NOT NULL,
+    vex_source TEXT NOT NULL DEFAULT '',
+    event_type TEXT NOT NULL DEFAULT 'falsePositive',
+    created_by_id TEXT NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
Hubtrick-Git, line 24: Are we actually using these timestamps anywhere?

database/migrations/20260601085124_add_system_vex_rules.up.sql
```diff
@@ -0,0 +1,36 @@
...
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    CONSTRAINT fk_system_vex_rules_cve
+        FOREIGN KEY (cve_id)
+        REFERENCES public.cves(cve)
+        ON DELETE CASCADE
+);
+
+-- Create indexes for better query performance
+CREATE INDEX IF NOT EXISTS idx_vex_rule_cve ON public.system_vex_rules(cve_id);
```
Hubtrick-Git: this index is obsolete since its a subset of the compound index. Compound indexes can also be used as subsets from left to right (in declaration order), so the compound index make the following indexes available:
 - cve_id 
 - cve_id, vex_source

daemons/system_vexrule_daemon.go
```diff
@@ -0,0 +1,72 @@
...
+		slog.Info("No SystemVEXRules in database yet, skipping SystemVEXRule application")
+		return nil
+	}
+	// Create VexRules from all AssetVersions and System vexrules
+	var applicableRules []models.VEXRule
+	for _, assetVersion := range assetVersions {
+		for _, systemVEXRule := range systemVEXRules {
+			// This VEXRule is only created temporarily for the execution of the daemon, because
+			// storing the VEXRule would interfere with the crowdsourced vexing
+			rule := models.VEXRule{
```
Hubtrick-Git: Why not use the transformer, you wrote, here?

daemons/system_vexrule_daemon.go
```diff
@@ -0,0 +1,72 @@
...
+	// Check paranoid mode or settings if auto apply is enabled (can be found in asset)
+	// Add application setting in asset model
+	// Change application setting when paranoid mode is changed
+	assetVersions, err = runner.assetVersionRepository.FindSystemVEXRuleApplicableAssetVersions(ctx, tx)
+	if err != nil {
+		slog.Error("failed to fetch assetVersions from database", "error", err)
+		return err
+	}
+
+	if len(assetVersions) < 1 {
```
Hubtrick-Git: len() == 0 ?

database/repositories/cve_relationship_repository.go
```diff
@@ -27,3 +28,43 @@ func (repository *cveRelationshipRepository) GetRelationshipsByTargetCVEBatch(ct
...
 	return relations, nil
 }
+
+func (repository *cveRelationshipRepository) FindCrossRelationshipsBatch(
+	ctx context.Context,
+	tx *gorm.DB,
+	associatedCVEIDs []string,
+) ([]models.CVERelationship, error) {
+
+	const chunkSize = 10000
```
Hubtrick-Git, line 45: why are we chunking here and why in chunks of 10_000 ?

database/repositories/dependency_vuln_repository.go
```diff
@@ -406,6 +406,36 @@ func (repository *dependencyVulnRepository) GetAllOpenVulnsByAssetVersionNameAnd
 
 }
 
+func (repository *dependencyVulnRepository) GetAllOpenVulnsByAssetVersionNameAndAssetIDBatch(
+	ctx context.Context,
+	tx *gorm.DB,
+	assetTuples []struct {
```
Hubtrick-Git: Does not look so clean to define a struct inside a function signature, also you define the same struct over and over again, I would define it once somewhere, export it and reuse it

normalize/sbom_graph.go
```diff
@@ -141,6 +142,41 @@ func NewVexReport(report *cdx.BOM, source string) (*VexReport, error) {
 	}, nil
 }
 
+type VexReportOpenVEX struct {
+	Report *ov.VEX
+	Source string
+}
+
+func validateVexReportOpenVEX(report *ov.VEX) error {
+	if report.ID == "" {
```
Hubtrick-Git: To avoid nil pointer dereferences you could also check if report is nil

services/crowdsourced_vexing_service.go
```diff
@@ -153,3 +175,39 @@ func (s *CrowdsourcedVexingService) Recommend(ctx shared.Context, tx shared.DB,
...
+outer:
+	for i := range validRules {
+		for j := range validRules {
+			if !s.cveRelationshipService.IsAlias(validRules[i].CVEID, validRules[j].CVEID, cveAliasMap) {
+				nonAliasDetected = true
+				break outer
+			}
+		}
+	}
+	if len(validRules) == 0 {
```
Hubtrick-Git: I'd move the check before the loop its confusing to check after the loop

services/cve_relationship_service.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+
+		cveAliasMap[rel.SourceCVE][rel.TargetCVE] = struct{}{}
+		cveAliasMap[rel.TargetCVE][rel.SourceCVE] = struct{}{}
+	}
+
+	return cveAliasMap, nil
+}
+
+func (s *CVERelationshipService) IsAlias(cveSource, cveTarget string, cveMap map[string]map[string]struct{}) bool {
```
Hubtrick-Git: just an idea but to avoid typing this long type of map[string]map[string]struct{}, you could define a new type for it and reuse that

services/scan_service.go
```diff
@@ -922,3 +927,123 @@ func (s *scanService) ScanSBOMWithoutSaving(ctx context.Context, bom *cyclonedx.
 		DependencyVulns: vulnDTOs,
 	}, nil
 }
+
+func (s *scanService) FetchOpenVexFromGitHub(ctx context.Context, targetURL string, targetBranch string) (vexReports []*normalize.VexReportOpenVEX, err error) {
```
Hubtrick-Git: Just architectural: maybe we can move this function to a different location. Maybe to an existing file or group it in something like scan_import or something like that. The other fetchVexfromUpstream function can also move there.

services/vex_rule_service.go
```diff
@@ -645,16 +828,110 @@ func matchRulesToVulns(rules []models.VEXRule, vulns []models.DependencyVuln) ma
...
+				systemVEXRule := transformer.VEXRuleToSystemVEXRule(parsedRule)
+				systemVEXRulesMap[parsedRule.ID] = true
+				systemVEXRules = append(systemVEXRules, systemVEXRule)
+				if _, cveExists := includedCVEsMap[parsedRule.CVEID]; !cveExists {
+					includedCVEs = append(includedCVEs, parsedRule.CVEID)
+				}
+			}
+		}
+	}
+	//Check if CVEs are already in database since database can take some time to be established
```
Hubtrick-Git: If you run into this problem use postgresql unnest feature. we used it a few times already on bulk operations. It circumvents the parameter limit by using 1 parameter per column and then just passing the values in separate arrays.

- Also i think your existing cves lookup map is very expensive in terms of memory, stick to struct{} as value type.

shared/common_interfaces.go
```diff
@@ -269,6 +269,14 @@ type DependencyVulnRepository interface {
 	// regardless of path. Used for applying status changes to all instances of a CVE+component combination.
 	FindByCVEAndComponentPurl(ctx context.Context, tx DB, assetID uuid.UUID, cveID string, componentPurl string) ([]models.DependencyVuln, error)
 	GetDirectDependencyFixedVersionByPackageName(ctx context.Context, tx DB, packageName string) (*string, error)
+	GetAllOpenVulnsByAssetVersionNameAndAssetIDBatch(
```
Hubtrick-Git: please keep the definition to 1 line to keep it clean and readable. The struct reuse of assetTuples also comes in handy here.

services/scan_service.go
```diff
@@ -922,3 +927,123 @@ func (s *scanService) ScanSBOMWithoutSaving(ctx context.Context, bom *cyclonedx.
...
+	resp, err := downloadRawFileFn(ctx, owner, repo, branch)
+	if err != nil {
+		return nil, err
+	}
+
+	repoZip, err := utils.ZipReaderFromResponse(resp)
+	if err != nil {
+		return nil, fmt.Errorf("could not read obtained zip: %w", err)
+	}
+	defer resp.Body.Close()
```
Hubtrick-Git: you can actually just outright close the body at this point no need to defer it and keep the connection dangling until the fetch is completed

services/scan_service.go
```diff
@@ -922,3 +927,123 @@ func (s *scanService) ScanSBOMWithoutSaving(ctx context.Context, bom *cyclonedx.
...
+		if !strings.HasSuffix(filename, ".json") {
+			continue
+		}
+
+		fileRead, err := fileEntry.Open()
+		if err != nil {
+			slog.Info("openvex document could not be opened, skipping this file for parsing", "filename", fileEntry.Name, "err", err)
+			continue
+		}
+		data, err := io.ReadAll(fileRead)
```
Hubtrick-Git: you can simplify and optimize this using json.Decoder on the fileRead Reader

Hubtrick-Git:
And one more general thing, try always to use code comments to roughly explain what is happening if its not obvious.
e.g. len() == 0 {return ...} is self-explanatory
but some logic like string splitting or long complex code blocks are better to understand with comments. 
Function comments explaining the job of a function also dont hurt

Hubtrick-Git (changes requested):

daemons/system_vexrule_daemon.go
Dboy0ZDev: The transformer will turn a systemvexrule into a vexrule "as it is".
In this loop we need to set some additional fields, so we can't use the transformer.

database/repositories/cve_relationship_repository.go
Dboy0ZDev: Arguments that are ingested and passed to the query exceed database query search argument limit of 65535. 10000 chunk size was chosen at random for testing and wasn't changed afterwards.

Hubtrick-Git: Ok thats understandable than I would also propose the unnest feature as a solution for this.

PR #2067, opened 2026-05-27
fix: retry Go license lookups with v prefix

### LINKED ISSUE #2058: license risk unknown, url has changed
https://github.com/protonmail/gopenpgp/v3
URL is MIT

https://deps.dev/go/github.com%2FProtonMail%2Fgopenpgp%2Fv3
Devguard is saying it is unknown

Maybe have a it scheduled and check if links are dead, or there is a license available if 404

iccccccccccccc:
﻿Fixes #2058

## Changes
- retry Go package lookups with a `v` prefix when deps.dev rejects the plain semver form
- keep the retry scoped to Go modules so other ecosystems keep their current request shape
- added a small regression test around the `gopenpgp/v3` style path

## Verification
- `go test ./services -run TestGetVersion -count=1`
- `go test ./services -run TestGetProject -count=1`
- `go test ./services -count=1`
- `git diff --check`

A quick manual check against deps.dev shows `github.com/ProtonMail/gopenpgp/v3` returns MIT for `v3.4.1`, while the same version without the leading `v` 404s. This keeps that case from falling back to `unknown` just because the imported PURL version is written as `3.4.1`.

### DIFF
--- a/services/open_source_insight_service.go
+++ b/services/open_source_insight_service.go
@@ -94,6 +94,30 @@ func (s *openSourceInsightService) GetVersion(ctx context.Context, ecosystem, pa
 		packageName = strings.ReplaceAll(packageName, "/", ":")
 	}
 
+	if ecosystem == "go" && version != "" {
+		return s.getGoVersion(ctx, ecosystem, packageName, version)
+	}
+
+	return s.getVersion(ctx, ecosystem, packageName, version)
+}
+
+func (s *openSourceInsightService) getGoVersion(ctx context.Context, ecosystem, packageName, version string) (dtos.OpenSourceInsightsVersionResponse, error) {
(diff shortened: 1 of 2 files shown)


timbastin:
Hi @iccccccccccccc thanks a lot for this PR. What do you think about changing the condition? So that for Golang, we first try with a "v" prefix-  or make sure that the provided version has one; if that fails, we try without it. Since the v prefix is a strong convention in go

iccccccccccccc:
Good point, yeah, that order makes more sense for Go. I updated it so Go versions are tried with the v prefix first, then fall back to the plain version if deps.dev doesn't have that one.

I also adjusted the tests to check the request order explicitly. Ran: go test ./services.

timbastin (approved):

PR #2107, opened 2026-06-09
umami telemetry data when starting up, init

5byuri:


cmd/devguard-scanner/commands/root.go
```diff
@@ -103,6 +108,10 @@ and 'attest' to interact with the platform. Configuration can be provided via a
 			return err
 		}
 
+		telemetryOnce.Do(func() {
+			telemetry.SendScannerStartup(cmd.Context(), telemetry.ConfigFromEnv(), nil, scannerVersion(), config.RuntimeBaseConfig.APIURL, runsInCI, cmd.Name())
+		})
```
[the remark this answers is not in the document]
5byuri: this is fixed now

timbastin:
- Pretty bloated, there are a lot of config options we just dont need
- We need a value for WebsiteID set
- Add a transparency log, when telemetry data is send to umami. Something like: "sending anonymized telemetry data - nothing personal or critical. This helps us understand what devguard versions are used and we should provide patches for. You can disable this by setting the following environment variable: ..."

telemetry/telemetry.go
```diff
@@ -0,0 +1,350 @@
...
+		Component:  ComponentScanner,
+		Version:    version,
+		InstanceID: HashParts(ComponentScanner, apiHost, goos, goarch),
+		Data:       data,
+	}
+}
+
+func SendAPIStartup(ctx context.Context, cfg Config, client HTTPDoer, statsCollector APIStatsCollector, version string) {
+	if cfg.Disabled {
+		SendStartup(ctx, cfg, client, APIStartupEvent(
```
timbastin: this looks odd right here. If the user decides not so send the telemetry, we should not send it at all.

telemetry/telemetry.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/l3montree-dev/devguard/database/models"
+	"github.com/l3montree-dev/devguard/shared"
+)
+
+const (
+	ComponentAPI     = "devguard-api"
+	ComponentScanner = "devguard-scanner"
+
+	EventName       = "devguard-instance-start"
+	DefaultEndpoint = "http://localhost:3001/api/send"
```
timbastin: This right here needs to be changed

timbastin (changes requested):

telemetry/telemetry.go
```diff
@@ -0,0 +1,344 @@
...
+)
+
+const (
+	ComponentAPI     = "devguard-api"
+	ComponentScanner = "devguard-scanner"
+
+	EventName         = "devguard-instance-start"
+	TelemetryEndpoint = "https://umami.l3montree.com/api/send"
+	WebsiteID         = "2ab9fe36-42ec-485d-a592-b0f6e78dd1ad"
+	DefaultTimeout    = 3 * time.Second
```
timbastin:
```suggestion
	DefaultTimeout    = 30 * time.Second
```

timbastin (approved):

PR #2118, opened 2026-06-10
improved guidance to setup the encryption key

Hubtrick-Git:
Just add more information on how to setup the encryption for the database.


.env.example
```diff
@@ -31,5 +31,9 @@ ALPINE_RELEASES_API=https://alpinelinux.org/releases.json
 CSAF_PASSPHRASE=example-passphrase
 DEPENDENCY_PROXY_BASE_URL=https://api.main.devguard.org/api/v1/dependency-proxy
 
-# key to handle encryption for secrets in the database, must be configured to start devguard
-APP_SIDE_ENCRYPTION_KEY_PATH=example_app_side_encryption_key.key
\ No newline at end of file
+# key to handle encryption for secrets in the database, MUST be configured to be able to start devguard:
+# To do so first generate a secure 256 bit aes key (in hexadecimal), then run:
```
seb-kw: Can you add a typical command to do so? Something like `openssl rand -hex 32 > app_side_encryption_key.key`. So that user do not have to worry how to create the key.

seb-kw (changes requested):

.env.example
Hubtrick-Git: Personally I would not do that because I think we do not want to be responsible for cryptographically secure key generation and every user should actively make this choice on their own.

seb-kw: We are actually already responsible for key generation (e.g. PATs) 😉

When it is formulated as 

```
To do so first generate a secure 256 bit aes key (in hexadecimal), e.g. with `openssl rand -hex 32 > app_side_encryption_key.key`...
```

then it is only a suggestion. User can always choose a different approach.

timbastin (approved):

PR #2119, opened 2026-06-10
Feature/pat expiry date

Hubtrick-Git:
## Backend

Adds an expiry date to all PATs.

### Discussion

1. Default expiry interval: I choose an arbitrary default of 14 days, this is definitely open for discussions.

2. Upper bounds for expiry dates: Similarly i choose an upper bound of approx. 1 year.

3. Clean up of expired tokens: Currently we do not clean up any expired tokens. This could potentially lead to a lot of      
    wasted data inside the table. To solve this we could:
        a.  use an explicit clean up command 
        b. lazy deletion: when we run into an expired token delete it. 

## Frontend

The backend now demands a time interval in which the pat is valid (in seconds). This mean when sending a request to create a PAT the frontend additionally needs to set the "expireAfterSeconds" JSON-key inside the request. It expects a positive integer in the range of 1 second to 1 year (in seconds).


controllers/pat_controller.go
```diff
@@ -59,6 +59,10 @@ func (p *PatController) Create(c shared.Context) error {
 		return echo.NewHTTPError(400, fmt.Sprintf("could not validate request: %s", err.Error()))
 	}
 
+	// expiry interval must exist, be positive and cannot exceed 1 year
+	if req.ExpireAfterSeconds <= 0 || req.ExpireAfterSeconds > 365*24*60*60 {
```
timbastin: I think we should use the validation struct tag for this. The request values are already getting validated in line 58. Checkout this example: https://github.com/go-playground/validator/blob/master/_examples/simple/main.go#L14

database/migrations/20260608111314_add_pat_expiry_date_column.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE public.pat
+  ADD COLUMN expiry_date TIMESTAMPTZ NOT NULL DEFAULT NOW() + INTERVAL '14 days';
```
timbastin:
```suggestion
  ADD COLUMN expiry_date TIMESTAMPTZ NOT NULL DEFAULT NOW() + INTERVAL '365 days';
```

Otherwise all PATs will expire super fast after releasing this change.

database/repositories/pat_repository.go
```diff
@@ -45,28 +46,23 @@ func (g *gormPatRepository) DeleteByFingerprint(ctx context.Context, tx *gorm.DB
...
-
+// checks if a valid token exists for the fingerprint, this excludes any expired tokens
 func (g *gormPatRepository) GetByFingerprint(ctx context.Context, tx *gorm.DB, fingerprint string) (models.PAT, error) {
 	var t models.PAT
 	err := g.GetDB(ctx, tx).First(&t, "fingerprint = ?", fingerprint).Error
+	if err != nil {
+		return t, err
+	}
+	// check if the token expired
+	if t.ExpiryDate == nil || t.ExpiryDate.Before(time.Now()) {
```
timbastin: From an architecture perspective I dislike the Expiry enforcement in the repository itself. What do you think about moving this if statement to the service?

timbastin (changes requested):

database/repositories/pat_repository.go
Hubtrick-Git: I had the same idea but in my opinion the existing code already treats a missing token as an error; so in that sense an error means that we cannot find a valid token to authenticate. Thats why I delegated the check into the same function. Also the obvious upside of this is that there is no need for duplicate code AND you cannot forget checking the expiry date on the token by accident (making it more resilient). what do you think?

database/migrations/20260608111314_add_pat_expiry_date_column.up.sql
Hubtrick-Git: yeah as stated above the default value is up for discussion, i was thinking that a low default value is more secure

database/repositories/pat_repository.go
timbastin: I think actually its different. Then we would need to rename the function at least to signal to the developer that this will check for expiry. I see the upside though, nevertheless, there might be use cases in the future which want to get an even expired pat by fingerprint, maybe just to check if it ever existed

Hubtrick-Git: Ok refactored it

timbastin (approved):

PR #2133, opened 2026-06-12
Cascade rbac rule deletion

Hubtrick-Git:
Now when an org or project or assets get deleted all associated RBAC rules get deleted as well.


controllers/org_controller.go
```diff
@@ -140,6 +140,11 @@ func (controller *OrgController) Delete(ctx shared.Context) error {
 		return echo.NewHTTPError(500, "could not delete organization").WithInternal(err)
 	}
 
+	// then delete all RBAC roles associated with this org as well
+	if err := controller.rbacProvider.RevokeAllRolesForDomain(organizationID); err != nil {
+		return echo.NewHTTPError(500, "could not delete organization roles").WithInternal(err)
```
seb-kw: "organization successfully deleted but could not delete organization roles"

accesscontrol/casbin_rbac.go
```diff
@@ -283,6 +284,53 @@ func (c *casbinRBAC) RevokeAllRolesInAssetForUser(ctx context.Context, user stri
...
+	if len(policiesToRemove) > 0 {
+		if _, err := c.enforcer.RemovePoliciesCtx(ctx, policiesToRemove); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *casbinRBAC) RevokeAllRolesInProject(ctx context.Context, project string) error {
+	return c.revokeAllRolesForPrefix(ctx, "project::"+project+"|")
```
seb-kw, line 365: Maybe add a comment here why the | is necessary?

controllers/delete_rbac_cleanup_test.go
seb-kw, line 1: May want to add a test case for `revokeAllRolesForPrefix`?

seb-kw (changes requested):

seb-kw (approved):

timbastin (changes requested):
I think we need a migration to delete all not used casbin rules as well. We have deployments which are already running since some time and those already accumulated a lot of non necessary rules

timbastin (approved):

PR #2136, opened 2026-06-15
reduces separate count queries by using postgresql window function

timbastin:


Hubtrick-Git:
Added a safeguard for empty pages (pagination) so they now return the actual count instead of 0.

Replaced find by scan, since scan does not preload all related tables properly like find does.

Isolated queries using gorm sessions.

database/repositories/first_party_vuln_repository.go
```diff
@@ -104,15 +101,32 @@ func (repository *firstPartyVulnerabilityRepository) GetByAssetVersionPaged(ctx
...
+	firstPartyVulns := make([]models.FirstPartyVuln, len(rows))
+	for i, r := range rows {
+		firstPartyVulns[i] = r.FirstPartyVuln
+		count = r.TotalCount
+	}
 
-	if err != nil {
-		return shared.Paged[models.FirstPartyVuln]{}, nil, err
+	// if we have no rows the window functions does not work, so we fallback to a traditional count
+	if len(rows) == 0 {
```
timbastin, line 126: if we have no rows, shouldnt the count just be 0?

Hubtrick-Git: in the case that we request an empty page the result would be 0 since there are no rows to append the total_count column to.

PR #2219, opened 2026-07-06
Add invitation status

Dboy0ZDev:
Issue: https://github.com/l3montree-dev/devguard/issues/889

Related to frontend merge: https://github.com/l3montree-dev/devguard-web/pull/803

- Implemented data aggregation to serve information about invitation status
- Implemented routes and revocation functionality for invitations
- Implemented Expiry check defaulted to 15 minutes for invitations

### DIFF
--- a/controllers/org_controller.go
+++ b/controllers/org_controller.go
@@ -31,6 +31,10 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+const (
+	expiryDuration = 48 * time.Hour
+)
+
 type OrgController struct {
 	organizationRepository shared.OrganizationRepository
 	orgService             shared.OrgService
@@ -38,6 +42,11 @@ type OrgController struct {
 	projectService         shared.ProjectService
 	invitationRepository   shared.InvitationRepository
 	adminService           shared.AdminService
+}
+
+func isInvitationExpired(invite models.Invitation) bool {
+	now := time.Now()
+	return now.After(invite.CreatedAt.Add(expiryDuration))
 }
 
 func NewOrganizationController(repository shared.OrganizationRepository, orgService shared.OrgService, rbacProvider shared.RBACProvider, projectService shared.ProjectService, invitationRepository shared.InvitationRepository, adminService shared.AdminService) *OrgController {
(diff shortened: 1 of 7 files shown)


controllers/org_controller.go
```diff
@@ -207,6 +210,12 @@ func (controller *OrgController) AcceptInvitation(ctx shared.Context) error {
 		return echo.NewHTTPError(404, "invitation not found").WithInternal(err)
 	}
 
+	// Expiry of invitation set to 15 min, we don't need to save it in the database if we adhere to convention like this
+	now := time.Now()
+	if now.After(invitation.CreatedAt.Add(15 * time.Minute)) {
```
timbastin:
```suggestion
	if now.After(invitation.CreatedAt.Add(48 * time.Hour)) {
```

controllers/org_controller.go
```diff
@@ -495,14 +504,35 @@ func (controller *OrgController) readDetails(ctx shared.Context) error {
...
+	invitations, err := controller.invitationRepository.FindByOrgID(ctx.Request().Context(), controller.organizationRepository.GetDB(ctx.Request().Context(), nil), organization.ID.String())
+	if err != nil {
+		return echo.NewHTTPError(500, "could not get invitations of organization").WithInternal(err)
+	}
+
+	var invitedUsers []dtos.InvitedUserDTO
+	now := time.Now()
+	for _, inv := range invitations {
+		var invitationStatus dtos.InvitationStatus
+		if now.After(inv.CreatedAt.Add(15 * time.Minute)) {
```
timbastin: Now the invitation expired duration (which I think would be better to be like 48 hours) is something like a magic number. We should have this as a constant or a helper function.

timbastin (changes requested):

controllers/org_controller.go
Dboy0ZDev: I moved it to its own separate function. Also since we consider frequent changes to this var, I also added that the expiry date information is send to the frontend so we can show the user the exact time when the invitation expires. https://github.com/l3montree-dev/devguard-web/pull/803/changes/1ff4af729b89feb398706f19c50c71a5f82e3c74

controllers/org_controller.go
```diff
@@ -495,14 +511,35 @@ func (controller *OrgController) readDetails(ctx shared.Context) error {
 	organization := shared.GetOrg(ctx)
 	// fetch the regular members of the current organization
 	members, err := shared.FetchMembersOfOrganization(ctx)
-
 	if err != nil {
 		return echo.NewHTTPError(500, "could not get members of organization").WithInternal(err)
 	}
 
+	invitations, err := controller.invitationRepository.FindByOrgID(ctx.Request().Context(), controller.organizationRepository.GetDB(ctx.Request().Context(), nil), organization.ID.String())
```
timbastin:
```suggestion
	invitations, err := controller.invitationRepository.FindByOrgID(ctx.Request().Context(), nil, organization.ID.String())
```

timbastin (approved):

PR #2338, opened 2026-07-10
feat: cache cve-ecosystem-distribution endpoint

### LINKED ISSUE #2222: Cache cve-ecosystem-distribution endpoint
Currently, requests to the cve-ecosystem-distribution endpoint are taking ages (about 15 to 20 s). As I would assume the data does not change often, maybe we can cache it.

domzoric:
## Summary

The `/api/v1/vulndb/cve-ecosystem-distribution/` endpoint takes 15-20s per request because it aggregates the whole global vuln db every time (`COUNT(DISTINCT cve_id)` over the full `cve_affected_component` join table).

The data only changes when the vulndb mirror runs (~hourly) and has no org/user/request dimension, so this adds a single in-process cached value with a 1h TTL and `singleflight` to prevent concurrent recomputations — the same pattern already used for the org statistics cache in `statistics_service.go`. The query logic itself is unchanged, just extracted into `computeCVEEcosystemDistribution`.

The cache hangs off the controller as a pointer so the existing value-receiver methods don't copy the lock.

## Tests

- Unit tests for the cache (empty/valid/expired) and proof that a cache hit never touches the database
- Integration test against a real Postgres (testcontainers): correct distribution incl. ecosystem lowercasing and merging of CVE + malicious package counts, and that a second call is served from cache
- `VulnDBController` was added to the fx `TestApp` wiring for this

Closes #2222

### DIFF
--- a/controllers/vulndb_controller.go
+++ b/controllers/vulndb_controller.go
@@ -1,10 +1,12 @@
 package controllers
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/l3montree-dev/devguard/config"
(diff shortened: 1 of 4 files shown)


timbastin (approved):

timbastin:
This is great!! Thank you @domzoric

PR #2457, opened 2026-07-13
Adds backend validation for min release age for dep proxy

seb-kw:
Related: https://github.com/l3montree-dev/devguard-web/pull/822


timbastin (changes requested):
Let's use struct tags for validation. We can create a custom DTO for this. This would fit in our current software architecture.

PR #2591, opened 2026-07-18
Fail health check when PostgreSQL listener disconnects

### LINKED ISSUE #2589: DevGuards health check should fail if database connection fails
This applies to the listen/notify strategy in casbin_watcher.go as well as regular database pool connections. I think just failing the health check is easier than implementing a reconnection strategy.

nicksan222:
## Summary

- mark the PostgreSQL broker unhealthy when a `LISTEN` connection fails
- include broker health in `/api/v1/health/`
- verify listener failure behavior against a real PostgreSQL instance

Closes #2589

### Latch listener failures

The broker's listener goroutine exits when `WaitForNotification` returns an error, and there is currently no reconnection strategy. Once that happens, the Casbin watcher can no longer receive policy change notifications.

A broker-wide atomic latch records this terminal failure. Health remains failed until the process is restarted, allowing the deployment's health-check policy to restore the listener without adding a partial reconnection mechanism.

This also avoids probing the listener connection from `IsHealthy`. The same PostgreSQL connection is already occupied by the blocking notification wait, so a concurrent ping does not reliably describe whether the listener loop is alive.

### Preserve database failure diagnostics

The health endpoint checks general database connectivity before checking the broker latch. This keeps the response specific:

- an unavailable database reports the existing database ping failure
- a reachable database with a terminated listener reports a pub/sub connection failure

### Keep recovery out of scope

The failed pooled connection is released as before. The change deliberately does not remove subscriptions or attempt to reconnect them; doing so would require preserving subscribers, re-running `LISTEN`, and defining retry behavior.

## Testing

- `go test ./database ./router ./accesscontrol`
- `go test ./tests -run '^TestBrokerIntegration$' -count=1`

The integration test establishes a real listener, terminates its PostgreSQL backend with `pg_terminate_backend`, and verifies that the broker becomes unhealthy.

### DIFF
--- a/database/postgresql_broker.go
+++ b/database/postgresql_broker.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log/slog"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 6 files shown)


timbastin:
Wow this looks great! Thanks a lot for this contribution!

PR #2600, opened 2026-07-21
feat: reworking on security advisory feature

### LINKED ISSUE #2590: Security Advisories Follow up
- I think we need to have some comments and the event stream. I noticed, that there are backend models for advisory events but they dont get used at all. Maybe we can just reuse the already established vuln_events and then reuse the RiskAssessmentFeed in the frontend. I think its pretty valuable to be able to add comments etc.
- The columns cannot be of semver type... For golang i have to add a v prefix. Either we adjust affected packages to not have a semver type column or we add a dedicated 'has_v_prefix' column, so that we are able to restore it correctly.
- I think we should have a created_at updated_at information displayed somewhere. I already started displaying it in the table but i think there is a timezone mismatch.
- There is a strange hover effect (text gets white) when publishing it

Image

juliankepka:


controllers/advisory_controller.go
```diff
@@ -156,3 +164,47 @@ func (controller *AdvisoryController) Delete(ctx shared.Context) error {
...
+		return echo.NewHTTPError(500, "could not get any data").WithInternal(err)
+	}
+
+	if advisory.AssetID != shared.GetAsset(ctx).ID {
+		return echo.NewHTTPError(404, "advisory not found")
+	}
+
+	userID := shared.GetSession(ctx).GetUserID()
+
+	var status dtos.DependencyVulnStatus
```
timbastin: I think we should rename this dto. I like that we are reusing it. Maybe we can even consolidate more dtos, basically a create event request dto and reuse it at all places

database/migrations/20260720115231_create_advisory_column_vulnevents.up.sql
```diff
@@ -0,0 +1,58 @@
...
+ALTER TABLE public.vuln_events ADD CONSTRAINT one_vuln_parent CHECK (
+  (dependency_vuln_id    IS NOT NULL)::int +
+  (license_risk_id       IS NOT NULL)::int +
+  (first_party_vuln_id   IS NOT NULL)::int +
+  (compliance_posture_id IS NOT NULL)::int +
+  (security_advisory_id  IS NOT NULL)::int = 1
+);
+
+ALTER TABLE public.affected_packages
+    ALTER COLUMN semver_introduced TYPE text USING semver_introduced::text,
```
timbastin, line 57: Since those aren't semver types anymore I think we should rename those columns to version_introduced and version_fixed.

database/models/advisory_model.go
```diff
@@ -4,10 +4,12 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/l3montree-dev/devguard/dtos"
 )
 
 type Advisory struct {
-	ID               int64             `json:"id" gorm:"primaryKey;column:id"`
+	Vulnerability    `gorm:"-"`
```
timbastin: What is the reason behind this? There should be some properties we can actually reuse. Is the struct so different?

database/models/advisory_model.go
```diff
@@ -17,13 +19,14 @@ type Advisory struct {
...
 	Visibility       string            `json:"visibility" gorm:"type:text;column:visibility;default:draft"`
+	Events           []VulnEvent       `json:"events" gorm:"foreignKey:SecurityAdvisoryID;constraint:OnDelete:CASCADE;"`
 }
 type AffectedPackage struct {
 	Model
 	Ecosystem        string     `json:"ecosystem" gorm:"type:text;column:ecosystem"`
 	PackageName      string     `json:"packageName" gorm:"type:text;column:package_name"`
-	SemverIntroduced *string    `json:"semverStart" gorm:"type:semver;index"`
-	SemverFixed      *string    `json:"semverEnd" gorm:"type:semver;index"`
+	SemverIntroduced *string    `json:"semverStart" gorm:"type:text;index"`
```
timbastin: Same here, I would prefer version introduced and version fixed

database/models/advisory_model.go
```diff
@@ -34,3 +37,43 @@ func (m Advisory) TableName() string {
...
+	switch state {
+	case dtos.VulnStatePublished:
+		m.Visibility = "public"
+	case dtos.VulnStateWithdrawn:
+		m.Visibility = "withdrawn"
+	}
+}
+
+func (m *Advisory) GetState() dtos.VulnState {
+	switch m.Visibility {
```
timbastin: What do you think about renaming the visibility to state? That would fit more naturally

services/advisory_service.go
```diff
@@ -53,3 +54,29 @@ func (s *AdvisoryService) Delete(ctx context.Context, tx shared.DB, id int64) er
 	}
 	return s.advisoryRepository.Delete(ctx, tx, id)
 }
+
+func (s *AdvisoryService) CreateVulnEventAndApply(ctx context.Context, tx shared.DB, userID string, advisory *models.Advisory, vulnEventType dtos.VulnEventType, justification string, mechanicalJustification dtos.MechanicalJustificationType, userAgent *string) (models.VulnEvent, error) {
```
timbastin, line 52: I think we can just move the implementation of the private method with the same name into this one. This is an unecessary indirection. I am sure you copied this pattern from some other part of the codebase. I would appreciate if you would do the same change there as well.

timbastin (changes requested):

database/models/advisory_model.go
juliankepka: AI made this change because i had errors with the embedding of the vulnerability. The justification was: The embedded Vulnerability struct is included for method promotion (Go embedding), so Advisory satisfies the shared vuln interface via methods like GetTicketID(), SetTicketID(), etc. — not for persistence. The gorm:"-" tag tells GORM to ignore this field entirely (no column mapping, no reads/writes, no migration). Without it, GORM would promote the embedded struct's fields (ID, AssetID, CreatedAt, UpdatedAt) into the advisories table, conflicting with Advisory's own identically-named columns and breaking auto-migration.

timbastin: We are not using auto migration. Let's try to align the structures. State is defined already on the vulnerability. You can checkout license risks for example and first party vulns.

database/models/advisory_model.go
```diff
@@ -16,15 +18,16 @@ type Advisory struct {
 	Severity         string            `json:"severity" gorm:"type:text;column:severity"`
 	VectorString     string            `json:"vectorString" gorm:"type:text;column:vector_string"`
 	AssetID          uuid.UUID         `json:"assetID" gorm:"type:uuid;column:asset_id"`
-	Visibility       string            `json:"visibility" gorm:"type:text;column:visibility;default:draft"`
+	State            string            `json:"state" gorm:"type:text;column:state;default:draft"`
```
timbastin: I think you can type this as dtos.VulnState already simplifying the other functions and maybe even reusing definitions from the vulnerability base struct

database/repositories/advisory_repository.go
```diff
@@ -92,12 +93,42 @@ func (advisoryRepository *AdvisoryRepository) Delete(ctx context.Context, tx *go
...
 func (advisoryRepository *AdvisoryRepository) GetAllAdvisoriesByAssetID(ctx context.Context, tx *gorm.DB, assetID uuid.UUID) ([]models.Advisory, error) {
 	advisories := []models.Advisory{}
-	public := "public"
+	// CSAF feed must contain published (public) AND withdrawn advisories - a
+	// withdrawn advisory stays part of the public record. Only drafts are excluded.
 	err := advisoryRepository.GetDB(ctx, tx).
 		Preload("AffectedPackages").
 		Where("asset_id = ?", assetID).
-		Where("visibility = ?", public).
+		Where("state IN ?", []string{statemachine.StatePublic, statemachine.StateWithdrawn}).
```
timbastin: Can we use dtos.VulnState here? 

Maybe it's simpler to filter for state != dtos.VulnStateDraft

database/repositories/advisory_repository.go
```diff
@@ -92,12 +93,42 @@ func (advisoryRepository *AdvisoryRepository) Delete(ctx context.Context, tx *go
...
 		Find(&advisories).Error
 	return advisories, err
 
 }
+
+func (advisoryRepository *AdvisoryRepository) ApplyAndSave(ctx context.Context, tx *gorm.DB, advisory *models.Advisory, vulnEvent *models.VulnEvent) error {
```
timbastin, line 108: You do not need to make that change but it might be worth the effort to establish a base vulnerability repository which provides such methods as generic functions to increase code reuse

dtos/dependency_vuln_dto.go
```diff
@@ -71,6 +71,8 @@ const (
 	VulnStateMarkedForTransfer VulnState = "markedForTransfer"
 	VulnStateImplemented       VulnState = "implemented"   // for compliance posture
 	VulnStateNotApplicable     VulnState = "notApplicable" // for compliance posture
+	VulnStatePublished         VulnState = "published"     // for advisory
+	VulnStateWithdrawn         VulnState = "withdrawn"     // for advisory
```
timbastin, line 76: I think draft is missing here? Or do we just use open?

dtos/dependency_vuln_dto.go
```diff
@@ -129,7 +131,7 @@ type DetailedDependencyVulnDTO struct {
 	Events []VulnEventDTO `json:"events"`
 }
 
-type DependencyVulnStatus struct {
+type CreateEventRequest struct {
```
timbastin, line 135: Love this!

transformer/advisory_transformer.go
```diff
@@ -45,19 +45,19 @@ func AdvisoryUpdateRequestToModel(c dtos.AdvisoryUpdate, advisory models.Advisor
 	if c.AssetID != nil {
 		advisory.AssetID = *c.AssetID
 	}
-	if c.Visibility != nil {
-		advisory.Visibility = *c.Visibility
+	if c.State != nil {
```
timbastin: This should only happen through an event to make sure we don't forget todo it

services/advisory_service.go
```diff
@@ -30,26 +31,48 @@ func (s *AdvisoryService) ReadAll(ctx context.Context, tx shared.DB, assetID uui
...
+func (s *AdvisoryService) ReadAdvisory(ctx context.Context, tx shared.DB, id uuid.UUID) (models.Advisory, error) {
 	return s.advisoryRepository.ReadAdvisory(ctx, tx, id)
 }
 
-func (s *AdvisoryService) Update(ctx context.Context, tx shared.DB, id int64, advisory *models.Advisory, currentVisibility string) error {
-	if currentVisibility != advisory.Visibility {
-		if err := statemachine.CheckStateTransition(currentVisibility, advisory.Visibility); err != nil {
-			return fmt.Errorf("invalid state change from %q to %q: %w", currentVisibility, advisory.Visibility, err)
+func (s *AdvisoryService) Update(ctx context.Context, tx shared.DB, id uuid.UUID, advisory *models.Advisory, currentState string) error {
+	if currentState != advisory.State {
```
timbastin: I think we should not update the state if not through the apply and save method

timbastin (changes requested):

database/repositories/advisory_repository.go
juliankepka: That is a good point. I would like to work on that but maybe on another branch so I can focus on the advisory for now.

dtos/dependency_vuln_dto.go
juliankepka: Yes, I'm using open so it just says "You created advisory"

PR #2626, opened 2026-07-22
Fix external project IDs after upsert

### LINKED ISSUE #2143: External-provider asset sync writes `project_id = nil UUID` (FK violation)
**Summary:** On re-sync, `syncProjectAssets` stamps every asset with `project.ID`, but for **already-existing** projects the upsert returns the struct with an unpopulated (`uuid.Nil`) ID. Assets are written against `00000000-…`, violating `fk_projects_assets`.

**Impact:** Assets fail to persist when re-syncing existing external projects (first import is fine). **163 occurrences**, clustered at sync runs.

**Evidence (prod):**
```
ERROR:  insert or update on table "assets" violates foreign key constraint "fk_projects_assets"
DETAIL: Key (project_id)=(00000000-0000-0000-0000-000000000000) is not present in table "projects".
```
All 163 are the zero UUID, all `UPDATE "assets"` (conflict-update branch).

**Root cause:**
1. Provider projects arrive with no DB id — [`upsertProjects`](services/external_entity_provider_service.go#L176) sets only `OrganizationID`.
2. [`UpsertSplit`](database/repositories/project_repository.go#L465) upserts on `(external_entity_provider_id, external_entity_id)`; conflict-update (existing) rows don't get their `ID` populated back, yet are returned in `updatedProjects`.
3. [`syncProjectAssets:306`](services/external_entity_provider_service.go#L306) trusts it: `assets[i].ProjectID = project.ID` (= `uuid.Nil`).

Always exactly the zero UUID, only on re-sync — not a race against a real project.

**Reproduce:**
```bash
docker exec -i devguard-postgresql-1 psql -U devguard -d devguard <<'SQL'
BEGIN;
UPDATE assets SET project_id = '00000000-0000-0000-0000-000000000000'
 WHERE id = (SELECT id FROM assets LIMIT 1);
ROLLBACK;
SQL
# => ERROR: ... violates foreign key constraint "fk_projects_assets"
#    DETAIL: Key (project_id)=(00000000-...0000) is not present in table "projects".
```
App-level: configure an external provider and run the project sync **twice**.

**Fix:** populate IDs for updated projects in `UpsertSplit` (re-select by external keys, map `id` back); guard `syncProjectAssets` to skip when `project.ID == uuid.Nil`.

mine-13-zoom:
## Summary

- reload externally managed projects after upsert and map their persisted database IDs back to both created and updated models
- stop asset synchronization before a zero project UUID can reach the asset upsert
- add unit and PostgreSQL integration coverage for the two-sync regression

## Testing

- `go test ./...`
- `go test ./database/repositories ./services`
- `go test ./tests -run '^TestProjectRepositoryUpsertSplitReturnsIDForUpdatedProject$' -count=1`
- `go vet ./database/repositories ./services`
- `golangci-lint run --new-from-rev=HEAD` (current container image)

Closes #2143

### DIFF
--- a/tests/project_repository_integration_test.go
+++ b/tests/project_repository_integration_test.go
@@ -1,0 +1,69 @@
+// Copyright (C) 2026 l3montree GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package tests
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/l3montree-dev/devguard/database/models"
+	"github.com/l3montree-dev/devguard/database/repositories"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestProjectRepositoryUpsertSplitReturnsIDForUpdatedProject(t *testing.T) {
+	db, _, terminate := InitDatabaseContainer("../initdb.sql")
+	defer terminate()
+
+	org := models.Org{Name: "External sync test", Slug: "external-sync-test"}
+	require.NoError(t, db.Create(&org).Error)
+
+	providerID := "gitlab"
+	externalEntityID := "group-123"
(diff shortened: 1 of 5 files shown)


tests/project_repository_integration_test.go
Hubtrick-Git, line 1: This test seems obsolete and with that maybe the whole fix. If you remove your changes in project_repository.go and then run the test it still passes. So either the test is not properly written or the mentioned problem stems from another bug (or maybe does not even exists anymore).

Hubtrick-Git (changes requested):

tests/project_repository_integration_test.go
mine-13-zoom: You are right. I copied this integration test unchanged onto a clean worktree at upstream/main (3d0931fc) and ran it against the PostgreSQL test container. It passes there as well: GORM already populates the fresh second-sync model with the persisted ID on the conflict-update path. That means the test does not reproduce #2143 and the repository reload is unnecessary. I am closing the PR rather than keeping a redundant fix. Thanks for catching this.

PR #2665, opened 2026-07-24
fix: new error code does not lead to instant error in ui

### LINKED ISSUE #2664: New Org shows Error instead of Empty Overview
Image

juliankepka:


controllers/statistics_controller.go
```diff
@@ -266,7 +266,7 @@ func (c *StatisticsController) GetOrgStatistics(ctx shared.Context) error {
 
 	orgStatistics, err := c.statisticsService.GetOrgStatistics(ctx.Request().Context(), org.ID, orgComponentsLimit, topCVEsLimit, topComponentsLimit, forceRefresh)
```
timbastin, line 267: Why does this error occur in the first place? Let's add an if condition before that statement.

controllers/statistics_controller.go
```diff
@@ -266,7 +266,7 @@ func (c *StatisticsController) GetOrgStatistics(ctx shared.Context) error {
...
 	if err != nil {
-		return echo.NewHTTPError(500, "could not calculate organization statistics").WithInternal(err)
```
Hubtrick-Git, line 269: we should not return 20x if we run into an actual error in the statistics function, we need to differentiate between those erros

Hubtrick-Git (dismissed):

Hubtrick-Git (approved):

timbastin (approved):

PR #2701, opened 2026-07-29
include mechanical justification in OpenVEX statements

refoo0:


services/asset_version_service.go
```diff
@@ -342,17 +344,25 @@ func dependencyVulnStateToImpactAnalysisState(state dtos.VulnState) cdx.ImpactAn
 	}
 }
 
-func getJustification(dependencyVuln models.DependencyVuln) *string {
+func getJustificationAndMechanicalJustification(dependencyVuln models.DependencyVuln) (*string, dtos.MechanicalJustificationType) {
+	var justification *string
+	var mechanicalJustification dtos.MechanicalJustificationType
 	// check if we have any event
 	if len(dependencyVuln.Events) > 0 {
 		// look for the last event which has a justification
```
Hubtrick-Git, line 352: Currently it actually retrieves the earliest justification. The justification variable gets overwritten on each matching event and since we are going backwards the earliest justification wins. If you only want the latest you are just missing the break statement.

services/asset_version_service.go
```diff
@@ -342,17 +344,25 @@ func dependencyVulnStateToImpactAnalysisState(state dtos.VulnState) cdx.ImpactAn
 	}
 }
 
-func getJustification(dependencyVuln models.DependencyVuln) *string {
+func getJustificationAndMechanicalJustification(dependencyVuln models.DependencyVuln) (*string, dtos.MechanicalJustificationType) {
+	var justification *string
+	var mechanicalJustification dtos.MechanicalJustificationType
```
Hubtrick-Git, line 349: since its not a pointer we are currently always returning (at least) an empty string, is that consistent with other implementations? If so we can keep it like that.

Hubtrick-Git (changes requested):

timbastin (approved):

PR #2733, opened 2026-08-02
Reduce unnecessary flake inputs and filter outputs

tfc:
This PR does:

- drop `flake-utils`
  - `flake-utils` are bigger an import than the LOC that are needed to get the same effect. Unnecessarily increased the size of the lock file and evaluation closure.
  - `flake-utils` make it intransparent what target platforms are actually used
- Filter outputs so that they don't appear for platforms where they don't really make sense. Mostly darwin.
- I also used `nixfmt` to format the changes.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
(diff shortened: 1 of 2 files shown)


tfc:
@timbastin ping

who is your local nix expert to discuss such topics with?

PR #2797, opened 2026-08-06
fix: fixed the Sync all sources button

juliankepka:


services/scan_service.go
```diff
@@ -756,65 +756,162 @@ func (s *scanService) FetchVexFromUpstream(ctx context.Context, assetID uuid.UUI
...
+	}
+
+	return s.IngestVEXRules(ctx, tx, asset, rules)
+}
+
+// IngestVEXRules diffs the given rules against the rules already stored for the asset - per VEX
+// source, so a rule that vanished from one source does not get removed just because another
+// source never contained it - and applies the newly added rules to the open vulnerabilities of
+// the asset.
+func (s *scanService) IngestVEXRules(ctx context.Context, tx shared.DB, asset models.Asset, rules []models.VEXRule) error {
```
timbastin: The scan controller already has this as a helper function. Maybe you can just declare it not as a method to reuse it. Besides that, lets try not to have a db access in NEW CODE we write in the service layer. That is a super new design philosophy, because we have trouble reusing code in two places: One as a controller method which operates on a single asset version, the other in the daemons which operate on multiple assets at once.

This results in a smaller service layer (basically only pure functions) and a bigger controller layer, doing all the database operations.

timbastin (changes requested):

timbastin (approved):

timbastin:
Looks great! Just linting and tests are failing. Might be worth to check, that we have a good test for this. Are vex rules correctly fetched and applied? I am not sure if a test already covers this.

PR #2821, opened 2026-08-10
continue syncing issues even if default asset version lookup fails

refoo0:


timbastin:
Maybe the reason for this PR was the issue, that when deleting the default asset version tickets wont get closed again

PR #2830, opened 2026-08-11
Improve fixed version daemon

Hubtrick-Git:
## Fixed Version Daemon

Changed the approach from a iterative approach processing 1 vuln at a time, to an aggregated processing. More concretely:

- Now filters for distinct dependency vulns instead of all vulns.
- Calculate fixed version in batches now instead of individual purls
- Updates rows using staging table instead of individual updates

## Purl Comparer Improvements

In Order to make the fixed version daemon faster, I also made some performance improvements regarding the affected components matching logic:

- The purl comparer now is customizable regarding what gets preloaded. The default is the existing preload behaviour, but fixed version daemon can skip all preloads except cves.
- The version based filter now happens BEFORE the additional preloads. SO we first fetch the colums necessary for version matching and then preload only the components which survived the filter
- Uses ANY instead of IN to for more efficient parameter parsing and query peformance

### Disclaimer

When testing against the main database it seems like no vulns need to be updated. After intensively searching for a bug I could not find one. It seems like that the underlying data is just up to date.


daemons/fixed_version_daemon.go
```diff
@@ -2,104 +2,230 @@ package daemons
...
 
-	slog.Info("updating fixed versions for dependency vulns", "count", len(dependencyVulns))
+// updates the dependency vulns passed using a <staging table + copy + update> from approach
+func (runner *DaemonRunner) UpdateAllFixedVersions(ctx context.Context, updatingJobs []fixedVersionJob) error {
+	if len(updatingJobs) == 0 { // fast exit if empty
+		return nil
+	}
 
-	wg := utils.ErrGroup[any](5)
+	if runner.pgxpool == nil {
```
timbastin: This is what some people (and especially AI) are calling defensive programming. I call it one of the worst mistakes one can make while programming.

If pgxpool really is nil, it MUST panic. If this condition really happens, WE AS DEVELOPERS did something wrong. It is not a condition that sometimes happens, and we should handle it gracefully. It is a invariant: "pgxpool HAS TO EXIST"; if not, panic and let me fix the bug.

Using this defensive programming, it will handle the condition gracefully, return an error, and log it to stdout just for us to never take a look at this. It would take much more time for us to find the error the developer did.

Hubtrick-Git: You are right regarding the code flow inside the trigger daemon function, since errors are not handled there but rather just loged. But if the caller (like admin controller) handles the error and fails/abort the job isnt that then the same outcome as a panic, since we recover from it?

This would still mean that a panic is better practice since with a panic you can be sure where and how it gets handled (parent recover function), whilst with errors you cannot be sure if they are swallowed or handled with the proper urgency.

timbastin (approved):

timbastin:
We can merge this as soon as the conflicts are resolved!

PR #2843, opened 2026-08-12
Add formatting, drop dead code

tfc:
This PR does:

- run statix (nix code analysis), run deadnix (drops dead code), nixfmt (formats nix code)
- adds treefmt, a meta-formatter that runs these 3 tools in 3 scenarios:
  - developer runs `nix fmt`
  - developer uses nix-shell and runs `treefmt`
  - CI runs `nix flake check` with the new github action and runs it as part of the checks

did not do that here but this leaves potential for:

- CI check could also push to the cache
- the treefmt config could also run go formatters/linters and other tools (markdown linters, etc. etc.)

This PR contains *no semantic changes*. Everything should stay untouched.

### DIFF
--- a/.github/workflows/nix-flake-check.yaml
+++ b/.github/workflows/nix-flake-check.yaml
@@ -1,0 +1,29 @@
+# Evaluates and builds all flake outputs (packages, devShells, checks)
+
+name: Nix-Flake-Check
+
+on:
+  workflow_dispatch:
+  push:
+
+permissions: {}
+
+jobs:
+  flake-check:
+    name: nix flake check
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - name: Check out code
+      uses: actions/checkout@f548e57e544e1ff5a4c46bf1e1b8685f8e4a348a # main @ 2026-07-20 (post v7.0.1)
+      with:
+        submodules: recursive
+        persist-credentials: false
+    - uses: cachix/install-nix-action@13d8dd58da0234aa297dedd986986ccb8e7f3e24 # v31.11.1
+      with:
+        extra_nix_config: |
+          substituters = https://cache.nixos.org https://nix.garage.l3montree.cloud
+          trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nix.garage.l3montree.cloud:MGlzfPQKA91/zxw91CN+GP7NpjAAwmKvWXlDYgeeI8k=
+    - name: Run nix flake check
+      run: nix flake check -L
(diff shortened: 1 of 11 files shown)


.github/workflows/nix-flake-check.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  flake-check:
+    name: nix flake check
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - name: Check out code
+      uses: actions/checkout@main
```
timbastin: I think we should pin the action to a specific commit to avoid any security issues

.github/workflows/nix-flake-check.yaml
```diff
@@ -0,0 +1,29 @@
...
+      with:
+        submodules: recursive
+        persist-credentials: false
+    - uses: cachix/install-nix-action@master
```
timbastin: Same here

flake.nix
```diff
@@ -61,9 +61,9 @@
         };
         # this is only done to satisfy the expected structure in the container hardening work
         binaries = import ./nix/devguard.nix {
-          buildGoModule = hostPkgs.buildGoModule;
-          lib = hostPkgs.lib;
-          inherit self system;
+          inherit (hostPkgs) buildGoModule;
```
timbastin: Can't we put that even in a single line with the one below?

flake.nix
```diff
@@ -99,9 +99,9 @@
         # Built for the evaluating system, so these are the only outputs that
         # mean anything on a non-Linux host.
         hostBinaries = {
-          devguardScanner = binaries.devguardScanner;
-          devguard = binaries.devguard;
-          devguardCLI = binaries.devguardCLI;
+          inherit (binaries) devguardScanner;
```
timbastin: Same here? Or am I wrong?

nix/oci.nix
```diff
@@ -1,21 +1,28 @@
...
+  uv2nix,
+  pyproject-build-systems,
+}:
+rec {
   devguardBinaries = import ./devguard.nix {
     inherit self;
-    buildGoModule = pkgs.buildGoModule;
-    lib = pkgs.lib;
+    inherit (pkgs) buildGoModule;
+    inherit (pkgs) lib;
```
timbastin: Same in this file

timbastin (changes requested):

flake.nix
tfc: Absolutely, thank you for pointing it out.

The reason how we got there is that statix can point out and fix `a = b.a;` patterns into `inherit (b) a;`, but it would not accumulate them in groups. I just ran the tool and then did not have a close look. Trustable output but arguably ugly, true.

tfc:
@timbastin all fixed, thank you for pointing these out.

timbastin (approved):

PR #2848, opened 2026-08-13
add org/project/asset slug label to GitHub/GitLab/Jira issues

refoo0:


timbastin (approved):

timbastin (changes requested):
Would be great to have a test for this functionality. You changed all tests to have an empty string as second parameter. Besides that, I think its a bit missleading, that we pass a label to a function which is called get label. What do you think about just passing all three slugs to GetLabels and that function calls commonint.SlugLabel

timbastin (approved):

PR #2854, opened 2026-08-14
fix: keep the stored createdAt when a webhook integration is updated

### LINKED ISSUE #2842: Updating Custom Webhook resets createdAt time to wrong set value
Before update through API call to `PUT /api/v1/organizations/{organization}/integrations/webhook/{id}/`
Image

After call to API:
Image

This error occurs on both organization scoped and project scoped webhooks

sujeito-operator:
Fixes #2842.

`WebhookController.Update` builds a fresh `models.WebhookIntegration` that carries only
the ID, so `Model.CreatedAt` is the zero time:

```go
webhookIntegration := &models.WebhookIntegration{
	Model: models.Model{
		ID: uuidID,
	},
	...
}
```

`webhookRepository.Save` ends in `GormRepository.Save`, which is `db.Save(t)`. For a
struct with a non-zero primary key gorm appends `Select("*")` — its own comment at
`finisher_api.go:112` (v1.31.2) reads *"when updating, use all fields including those
zero-value fields"* — so `SelectAndOmitColumns` marks every column and
`ConvertToAssignments` emits `created_at` too, zero value and all.

The statement gorm builds for that model, postgres dialector, `DryRun`:

```sql
UPDATE "webhook_integrations" SET "created_at"=$1,"updated_at"=$2,"name"=$3,"description"=$4,"url"=$5,"secret"=$6,"sbom_enabled"=$7,"vuln_enabled"=$8,"org_id"=$9,"project_id"=$10 WHERE "id" = $11
```

with `$1` bound to `0001-01-01 00:00:00 +0000 UTC`. Carrying the stored timestamp over gives the identical
statement with `$1` bound to `2026-08-12 13:59:38 +0000 UTC`. Both the organisation- and the
project-scoped route reach this one handler, which is why it happens on both.

The handler already reads the row it is about to overwrite — `GetClientByIntegrationID`,
for `OrgID` and `ProjectID` — so the timestamp comes from the read that is already there.
Two fields on one struct literal; no other behaviour changes.

`controllers/webhook_controller_test.go` is new and pins the behaviour where it broke rather than where it shows up:
it captures the model handed to the repository and asserts the timestamp survived.
Against `main` it fails with

```text
expected: 2026-08-12 13:59:38 +0000 UTC
actual  : 0001-01-01 00:00:00 +0000 UTC
```

and passes with the patch. `go test ./controllers/...` is green.

### One thing I found next to this and deliberately did not touch

The same handler assigns `Secret: &data.Secret` unconditionally, and no response DTO
carries the secret — `dtos.WebhookIntegrationDTO` has `id`, `name`, `description`, `url`,
`sbomEnabled`, `vulnEnabled` and nothing else. A client cannot send back a value it was
never given, so a `PUT` that omits `secret` replaces the stored signing secret with the
empty string, and `services.NewWebhookService` then sets an empty `X-Webhook-Secret` on
every later delivery. Whether an absent `secret` should mean *unchanged* or *cleared* is
a decision rather than a patch, so this pull request leaves it alone; it seemed worth an
issue of its own.

Written by an automated agent. Every figure above is produced by the script that opened
this pull request, which re-runs the measurement and refuses to open if anything drifts.


Hubtrick-Git:
Hey @sujeito-operator, thank you for your contribution, I requested some minor changes to your pr. When they are fixed we should be good to merge

controllers/webhook_controller.go
```diff
@@ -106,9 +106,12 @@ func (w *WebhookController) Update(ctx shared.Context) error {
 		return ctx.JSON(500, "failed to get webhook integration")
 	}
 
+	// Save() writes every column, including the zero ones, so the stored createdAt has to be
```
Hubtrick-Git: Lets shorten the comment into one line and move it next to the value assignment of created_at inside the struct

controllers/webhook_controller_test.go
Hubtrick-Git, line 1: The test is a good PoC but I would remove it to reduce our test volume, since it does not carry a lot of importance going forward from this pr

Hubtrick-Git (dismissed):
Good job, this definitely fixes the bug. Just some minor changes need to be made

controllers/webhook_controller.go
sujeito-operator: Done in d511a6e — one line, sitting on the `CreatedAt` assignment itself.

controllers/webhook_controller_test.go
sujeito-operator: Removed in d511a6e. Fair call — the behaviour it pinned is one line, and the comment now says why.

sujeito-operator:
Both done in d511a6e — comment shortened onto the assignment, `controllers/webhook_controller_test.go` deleted. The diff is one file now, +1/-1. `go build ./...`, `go vet ./controllers/...` and `gofmt` are clean on the branch.

Thanks for the quick read.

Hubtrick-Git (approved):

PR #2859, opened 2026-08-14
fix: retry SaveBatchBestEffort with smaller batches on protocol parameter limit

refoo0:


database/repositories/gorm_repository.go
Hubtrick-Git, line 1: I do not really understand the goal of this change:

The comments alone are contradictory to the idea of saveBatchBestEffort. We cannot go below 1 so the current function would already try to save each one on its own. Therefore these changes do not serve any benefit really. You are basically running the same save operation again and then just swallowing any errors instead of handling them like the existing if statement below. 

Besides that, if I am not missing something, there are several errors in the code itself:

- If the save (before your if statement) succeeds you just roll it back instantly, resulting in no changes at all, leading the function to always keep splitting down to 1 and then returning nil (with no changes)
- Your splitting logic at the end of your if statement only splits half the entries the other half is discarded.
- The code branch is newBatchSize < 1 cannot be reached

Hubtrick-Git (changes requested):

PR #2860, opened 2026-08-14
allow generating CSAF report with zero vulnerabilities

refoo0:


services/csaf_service.go
```diff
@@ -524,6 +520,9 @@ func (service csafService) GenerateCSAFReportForVulns(ctx context.Context, orgNa
 			URL:              new("https://first.org/tlp"),
 		},
 	}
```
Hubtrick-Git, line 522: Before returning the CSAF report make sure that the property document.Category is set. Its a mandatory category and therefore need to be set on every valid CSAF report. In Line L#555 this is explicitly done, you can just reuse this logic and move it up.

Hubtrick-Git (changes requested):

timbastin (approved):

timbastin (approved):

PR #2881, opened 2026-08-19
feat: added per-asset exposure metrics

### LINKED ISSUE #1985: Add Exposure to an Artifact
**Is your feature request related to a problem? Please describe.**
I would like to improve the score / accuracy of findings by classifying if an artifact is exposed over the internet.

**Describe the solution you'd like**
Add an exposure field to artifacts with a customizeable list of values (e.g. internet = high, intranet = medium, offline = none) which have influence on the CVSS score within devguard.

### LINKED ISSUE #2873: Add per-asset exposure metrics (MAV/MAC/MPR/MS) to risk formula
Risk is currently (CVSS-BE × (EPSS + 1)) / 2 / ComponentDepth, using each finding’s Base vector. CR/IR/AR (CIA assessment) is already captured per asset. 

Missing: exposure/reachability context — Base AV assumes a generic deployment, so identical CVEs on differently-exposed assets score identically.

Proposal: add asset-level fields, applied as CVSS Environmental overrides on the finding’s Base vector:

- MAV — network exposure: N/A/L/P
- MAC — compensating controls (WAF/segmentation): L/H
- MPR — baseline privilege to reach asset: N/L/H
- MS — lateral-movement/blast-radius potential: U/C

At score time: merge asset’s {MAV,MAC,MPR,MS} + existing CR/IR/AR with finding’s Base vector → CVSS Environmental score → feed into existing formula in place of CVSS-BE. Fields optional; unset = fall back to current Base-only behavior.

Questions for the settings page:

MAV — Is the asset reachable from the public internet, internal network only, a restricted segment, or is it airgapped/physical-access-only?

MAC — Are there compensating controls in front of it (WAF, segmentation, rate limiting) that make exploitation harder than the generic case?

MPR — What’s the minimum privilege needed just to reach the asset (anonymous, authenticated, MFA-gated)?

MS — If this asset is compromised, could it be used to pivot to other systems (shared credentials, same host/cluster, broad IAM permissions)?

juliankepka:


database/migrations/20260819075101_adjustAssetTable.up.sql
Hubtrick-Git, line 1: I would use NULL as the default value, safes memory and is more straightforward than an arbitrary default value.

database/models/asset_model.go
```diff
@@ -31,23 +31,30 @@ type AssetTuple = struct {
...
+	Avatar                       *string                         `json:"avatar" gorm:"type:text"`
+	Slug                         string                          `json:"slug" gorm:"type:text;uniqueIndex:idx_app_project_slug;not null;"`
+	ProjectID                    uuid.UUID                       `json:"projectId" gorm:"uniqueIndex:idx_app_project_slug;not null;type:uuid;"`
+	Description                  string                          `json:"description" gorm:"type:text"`
+	Type                         AssetType                       `json:"type" gorm:"type:text;not null;"`
+	AssetVersions                []AssetVersion                  `json:"refs" gorm:"foreignKey:AssetID;references:ID;constraint:OnDelete:CASCADE;"`
+	Project                      Project                         `json:"project" gorm:"foreignKey:ProjectID;references:ID;constraint:OnDelete:CASCADE;"`
+	Importance                   int                             `json:"importance" gorm:"default:1;"`
+	ConfidentialityRequirement   dtos.RequirementLevel           `json:"confidentialityRequirement" gorm:"default:'high';not null;type:text;"`
```
Hubtrick-Git: At some point you already define a struct to group all these attributes inside one object. Move the definition into here and reuse where ever you deal with all of the metrics at once.

dtos/asset_dto.go
```diff
@@ -14,6 +14,58 @@ const (
 	RequirementLevelHigh   RequirementLevel = "high"
 )
 
+type ModifiedRequirementLevel string
+
+const (
```
Hubtrick-Git, line 19: Small inconsistency: you are not using shortforms for this one only

dtos/asset_dto.go
```diff
@@ -92,12 +151,19 @@ type AssetCreateRequest struct {
...
 	CentralDependencyVulnManagement bool `json:"centralDependencyVulnManagement"`
 
-	Importance            int  `json:"importance"`
-	ReachableFromInternet bool `json:"reachableFromInternet"`
+	Importance int `json:"importance"`
 
 	ConfidentialityRequirement string  `json:"confidentialityRequirement" validate:"required,oneof=low medium high"`
 	IntegrityRequirement       string  `json:"integrityRequirement" validate:"required,oneof=low medium high"`
 	AvailabilityRequirement    string  `json:"availabilityRequirement" validate:"required,oneof=low medium high"`
+	ModifiedAttackVector       string  `json:"modifiedAttackVector" validate:"omitempty,oneof=X network adjacent local physical"`
```
Hubtrick-Git: If you already defined your own type for these attributes use them wherever you work with them. That will be dealt with when you unify them all into a single struct.

dtos/asset_dto.go
```diff
@@ -37,10 +89,17 @@ type AssetDTO struct {
 	Description string    `json:"description"`
 	ProjectID   uuid.UUID `json:"projectId"`
 
-	AvailabilityRequirement    RequirementLevel `json:"availabilityRequirement"`
-	IntegrityRequirement       RequirementLevel `json:"integrityRequirement"`
-	ConfidentialityRequirement RequirementLevel `json:"confidentialityRequirement"`
-	ReachableFromInternet      bool             `json:"reachableFromInternet"`
+	AvailabilityRequirement    RequirementLevel           `json:"availabilityRequirement"`
```
Hubtrick-Git: I am pretty sure you are better of using pointers for these attributes, since they are probably optional. Will also be dealt with once you unify them all into a single struct. 

Generally speaking if an attribute is optional you make it a pointer so that go can treat is as nil.

shared/context_utils.go
Hubtrick-Git: this is the struct i am talking about, use the right types and pointers and move it into the model definition

shared/context_utils.go
```diff
@@ -833,35 +833,60 @@ type Environmental struct {
...
-			return newValue
+	reqToShort := map[string]string{"low": "L", "medium": "M", "high": "H"}
+	mavToShort := map[string]string{"network": "N", "adjacent": "A", "local": "L", "physical": "P", "X": "X"}
+	macToShort := map[string]string{"low": "L", "high": "H", "X": "X"}
+	mprToShort := map[string]string{"none": "N", "low": "L", "high": "H", "X": "X"}
+	msToShort := map[string]string{"unchanged": "U", "changed": "C", "X": "X"}
+	muiToShort := map[string]string{"X": "X", "none": "N", "required": "R"}
+	mreqToShort := map[string]string{"X": "X", "none": "N", "low": "L", "high": "H"}
+
+	lookup := func(m map[string]string, v string) string {
```
Hubtrick-Git: i know it was already there before, but does it make sense to just use the value if we could not map it to a valid one? Does it make sense to use a default here?

transformer/asset_transformer_test.go
```diff
@@ -38,21 +38,6 @@ func TestApplyToModel(t *testing.T) {
 			},
 			updated: true,
 		},
-		{
```
Hubtrick-Git, line 41: you removed the test case for the attribute you removed, but did not add test cases for the ones you added

Hubtrick-Git (changes requested):
Overall a good job, just some minor design changes. Most prominently :

- Moving all the new attributes inside a separate struct with correct types and pointers
- Changing the default value of the attributes to NULL and nil respectively
- Adjust the tests

database/migrations/20260819075101_adjustAssetTable.up.sql
timbastin: This "X" isnt a default value but has the "nothing set" meaning in cvss calculation. Its a tradeoff: Storage vs simpler code, since you can always be sure to get correct values in those asset properties. If you dont mind @Hubtrick-Git, and I'm sure you agree that assets are not the primary storage factor, we'll just go with this since there are no "hard feelings" against it.

controllers/asset_controller.go
```diff
@@ -273,6 +273,14 @@ func (a *AssetController) Read(ctx shared.Context) error {
 	return ctx.JSON(200, transformer.AssetModelToDetailsDTO(app, members))
 }
 
+func applyChange[T ~string](changes *[]string, label string, field *T, patch *T) {
```
timbastin, line 276: this is beautiful!

shared/context_utils.go
timbastin: This doesnt make sense. We should map to the default value "X"

timbastin: Besides that, we can even move the maps outside of the function definition since they are static

database/migrations/20260819075101_adjustAssetTable.up.sql
Hubtrick-Git: Of course storage is irrelevant here, its more about semantics IMO.
If X is a valid choice you can make (meaning i am choosing to set it to undefined or not set) then I totally agree with you. But if each asset fits into exactly 1 of the categories and X is just a placeholder for "currently unknown" then NULL is semantically the perfect fit here.

juliankepka: Bildschirmfoto 2026-08-21 um 14 43 14

Bildschirmfoto 2026-08-21 um 14 45 18


Just to give you a better overview. There exposure metrics are all optional and can be adjusted in the asset settings. They are all set to X because IMO it is irrelevant if the DB saves X or notDefined. Both implicate the Not Defined state and first.org names it X as well. I would not suggest to set the default to NULL because that would be more confusing.

dtos/asset_dto.go
juliankepka: I understand your point but I think it is not necessary here because all the fields get a default X value so there is no possible outcome where a pointer would prevent a nil value. At a PATCH Request I absolutely understand your suggestion.

PR #2888, opened 2026-08-19
feat: sign webhook payloads with HMAC

### LINKED ISSUE #2206: Add HMAC signing for webhook payloads
Right now when a webhook has a secret configured, it gets sent as a plain X-Webhook-Secret header. Anyone who can see the request can grab the secret and reuse it. The receiver also has no way to verify the body hasn't been modified between DevGuard and their endpoint.

Compute an HMAC-SHA256 of the JSON body using the webhook secret, and send it as an X-Hub-Signature-256 header; similar to how GitHub and GitLab sign their webhook payloads. Receivers can then recompute the HMAC on their end to confirm the payload is authentic and intact. The old X-Webhook-Secret header can stay for anyone relying on it.

alternatives you've considered
 - Letting receivers verify via the existing X-Webhook-Secret header alone, but that doesn't prove the body wasn't altered.
 - Using a different hash algorithm; SHA256 is standard, well-supported, and what most existing webhook receivers expect.
 - Making this opt-in via a new config field, but signing should just happen automatically when a secret is present.

### Happy to implement this if it sounds useful.

amrrdev:
## Summary

  Fixes #2206
  Add HMAC-SHA256 signing for outgoing webhook payloads.

  ## Changes

  - Generate an HMAC-SHA256 signature from the exact JSON request body using the configured webhook secret.
  - Send the signature using the X-Hub-Signature-256 header.
  - Preserve the existing X-Webhook-Secret header for backward compatibility.
  - Skip the HMAC header when no secret is configured.
  - Reuse the same signature across webhook retries.
  - Add tests for signature generation, missing secrets, legacy header support, and retry consistency.

  ## Testing

  go test -v ./services -run 'TestWebhookClient_CreateRequest' -count=1

  All webhook tests pass.

### DIFF
--- a/services/webhook_service.go
+++ b/services/webhook_service.go
@@ -100,7 +100,7 @@ func (c *webhookClient) CreateRequest(ctx context.Context, method, url string, b
 		if err != nil {
 			return nil, err
 		}
-		if c.Secret != nil {
+		if c.Secret != nil && *c.Secret != "" {
 			req.Header.Set("X-Webhook-Secret", *c.Secret)
 		}
 		if signature != "" {
--- a/services/webhook_service.go
+++ b/services/webhook_service.go
@@ -6,6 +6,9 @@ package services
 import (
 	"bytes"
 	"context"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


timbastin (approved):

timbastin:
This looks great! Thanks a lot @amrrdev !

PR #2895, opened 2026-08-20
fix: manage artifacts now has duplicated key error

### LINKED ISSUE #2892: "Artifact name is already taken" is not good communicated.
Image

It should not just state "Internal Server Error"

juliankepka:


Hubtrick-Git (approved):
optional remove the other rollbacks as well

timbastin (approved):

PR #2900, opened 2026-08-21
fix: use CycloneDX 1.6 for sample SBOM

### LINKED ISSUE #2898: CycloneDX SBOM v 1.4 is being send instead of v 1.6 in test webhook
When sending a test evenet type SBOM over a Webhook i get a CycloneDX v 1.4 SBOM instead of a v 1.6

I used the https://webhook.site/ and the DevGuard Testprojects group from the for testing.

Test payload 

Image

It should look like this (taken from DevGuard-Web)

Image

henriquejsza:
## Summary

- emit sample SBOM webhook payloads using CycloneDX 1.6
- add focused regression coverage for the sample payload's spec version

## Validation

- `go test ./services -count=1`
- all non-Docker packages from the repository test command passed
- `golangci-lint run --new-from-rev=upstream/main` (0 issues)
- `go vet ./services`
- `gofmt` and `git diff --check`

The root integration-test package requires Docker/Testcontainers, which is not available in the local environment. Its Docker-provider failure reproduces unchanged on pristine `upstream/main`.

Closes #2898

### DIFF
--- a/services/webhook_service.go
+++ b/services/webhook_service.go
@@ -209,7 +209,7 @@ func testPayload(payloadType TestPayloadType) (any, WebhookType) {
 func createSampleSBOM() cdx.BOM {
 	return cdx.BOM{
 		BOMFormat:    "CycloneDX",
-		SpecVersion:  cdx.SpecVersion1_4,
+		SpecVersion:  cdx.SpecVersion1_6,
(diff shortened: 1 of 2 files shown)


timbastin (approved):

timbastin:
@henriquejsza Thanks a lot!

PR #2909, opened 2026-08-24
Improved api reference for custom webhooks

Dboy0ZDev:
In accordance to https://github.com/l3montree-dev/devguard-documentation/pull/243/ and https://github.com/l3montree-dev/devguard/issues/2847

Adjusted code annotations to reflect request bodies that are needed for custom webhook endpoints.


controllers/webhook_controller.go
```diff
@@ -18,6 +18,28 @@ import (
 type WebhookController struct {
 	webhookRepository shared.WebhookIntegrationRepository
 }
+type webhookUpdateData struct {
```
timbastin: can we move those into the dtos package?

timbastin (changes requested):

controllers/webhook_controller.go
Dboy0ZDev: Done 👍

timbastin (approved):

PR #2925, opened 2026-08-26
Eliminate pg protocol limit error sources

Hubtrick-Git:
- Added semgrep rule to identify possible SQL where we could hit the parameter limit most likely


database/repositories/gorm_repository.go
```diff
@@ -104,10 +107,18 @@ func (g *GormRepository[ID, T]) All(ctx context.Context, tx *gorm.DB) ([]T, erro
 	return ts, err
 }
 
-func (g *GormRepository[ID, T]) DeleteBatch(ctx context.Context, tx *gorm.DB, m []T) error {
-	err := g.GetDB(ctx, tx).Delete(m).Error
-	if err != nil {
-		return err
+func (g *GormRepository[ID, T]) DeleteBatch(ctx context.Context, tx *gorm.DB, entries []T) error {
```
Hubtrick-Git, line 110: Delete only needs the primary key columns as parameters so the batchSize could be optimized (on average prob at least 5x bigger batches) regarding delete queries, need to be discussed if thats worth it or not.

timbastin (approved):

PR #2928, opened 2026-08-27
feat(oci): add tar binary to postgresql container image

### LINKED ISSUE #2924: postgresql restore - `kubectl cp` command requires `tar` binary
**Is your feature request related to a problem? Please describe.**
I tried to use the documented way of restoring a `postgresql` backup using the `kubectl cp` command ([docs link](https://docs.devguard.org/how-to-guides/administration/backup-restore/#restore)), but that fails because the postgresql container does not contain the `tar` binary required by the `kubectl cp` command ([see output example](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_cp/#examples) from kubectl docs).

**Describe the solution you'd like**
I would like to extend the container image to contain the `tar` binary, so that restoring backups will be as simple as stated by the documentation.

**Describe alternatives you've considered**
An alternative would be do create a port-forward to the running postgresql pod and connect to it with a (local) psql command. This is harder to set-up though and kubectl port-forwarding is sometimes pretty unreliable (in my experience).

**Additional context**
I am not super familiar with nix, but my guess would be to simply add `pkgs.gnutar` to the list of the docker content here:
https://github.com/l3montree-dev/devguard/blob/d3d3ed7126baf7ca3c9084474fb780b6b83b5f20/nix/oci.nix#L212-L226

If that is the case, I could easily open a PR if that helps you.
Thanks in advance

skuethe:
This PR will add the `tar` binary (using the `gnutar` nix package) to the `devguard-postgresql` container image.
The tar binary will enable the usage of the `kubectl cp` command. Taken from the command output:

```bash
$ kubectl cp --help
Copy files and directories to and from containers.

Examples:
  # !!!Important Note!!!
  # Requires that the 'tar' binary is present in your container
  # image.  If 'tar' is not present, 'kubectl cp' will fail.
[...]
```
I haven't found any docs or tests related to the OCI build details which should also be updated.
If I missed to add something, please let me know.

closes #2924

### DIFF
--- a/nix/oci.nix
+++ b/nix/oci.nix
@@ -222,6 +222,7 @@ rec {
         postgresql.config
         pkgs.bash
         pkgs.coreutils
+        pkgs.gnutar
       ]
       ++ (if debug then [ pkgs.busybox ] else [ ]);
 


timbastin (approved):

timbastin:
Thanks a lot!

PR #2929, opened 2026-08-27
Implemented baseURL endpoint through instance settings

Dboy0ZDev:
Added that the instance-settings endpoint also returns information about the gitlaboauth2 configs.

This is in accordance with https://github.com/l3montree-dev/devguard-web/pull/914

### DIFF
--- a/controllers/admin_controller.go
+++ b/controllers/admin_controller.go
@@ -33,6 +33,7 @@ import (
 	"github.com/l3montree-dev/devguard/integrations/gitlabint"
 	"github.com/l3montree-dev/devguard/monitoring"
 	"github.com/l3montree-dev/devguard/shared"
+	"github.com/l3montree-dev/devguard/transformer"
 	"github.com/l3montree-dev/devguard/utils"
 	"github.com/labstack/echo/v4"
 	"gorm.io/gorm"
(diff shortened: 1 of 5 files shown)


controllers/admin_controller.go
```diff
@@ -378,9 +382,18 @@ func (controller *AdminController) GetInstanceSettings(ctx shared.Context) error
...
 	if err != nil {
-		return ctx.JSON(200, shared.InstanceSettings{})
+		return ctx.JSON(200, dtos.InstanceSettingsDTO{})
+	}
+	allGitlabOAuthConfigs := []dtos.GitlabOauth2ConfigDTO{}
+	for _, config := range controller.gitlabOAuth2 {
+		allGitlabOAuthConfigs = append(allGitlabOAuthConfigs, dtos.GitlabOauth2ConfigDTO{ProviderID: config.ProviderID, GitlabBaseURL: config.GitlabBaseURL})
 	}
-	return ctx.JSON(200, settings)
+	allGitlabOAuthConfigs = append(allGitlabOAuthConfigs, dtos.GitlabOauth2ConfigDTO{ProviderID: "gitlab", GitlabBaseURL: "http://localhost/test"})
```
timbastin: I think this right here was only for testing. Maybe we can move the transformation process into the transformers package. What do you think?

timbastin (changes requested):

controllers/admin_controller.go
Dboy0ZDev: Removed the test. Best I could do with the transformer was to move the object creation out. Moving the loop creates a circular import error.

controllers/admin_controller.go
```diff
@@ -378,9 +383,15 @@ func (controller *AdminController) GetInstanceSettings(ctx shared.Context) error
 func (controller *AdminController) InstanceSettingsPublic(ctx shared.Context) error {
 	settings, err := controller.configService.GetInstanceSettings(ctx.Request().Context())
 	if err != nil {
-		return ctx.JSON(200, shared.InstanceSettings{})
+		return ctx.JSON(200, dtos.InstanceSettingsDTO{})
+	}
+
+	gitlabOAuthConfigs := make([]dtos.GitlabOauth2ConfigDTO, 0, len(controller.gitlabOAuth2))
+	for _, config := range controller.gitlabOAuth2 {
+		gitlabOAuthConfigs = append(gitlabOAuthConfigs, dtos.GitlabOauth2ConfigDTO{ProviderID: config.ProviderID, GitlabBaseURL: config.GitlabBaseURL})
```
timbastin, line 391: here is still the transformation happening - can we move that inside the transformer as well or do we get a circular dependency?

timbastin: just tested it - we get an import cycle

timbastin (approved):

PR #3013, opened 2026-09-02
fix: ingest of vex now working from external sbom url

juliankepka:


services/scan_service.go
```diff
@@ -975,3 +979,30 @@ func (s *scanService) ScanSBOMWithoutSaving(ctx context.Context, bom *cyclonedx.
...
+		}
+	}
+
+	if len(externalURLs) == 0 {
+		return nil
+	}
+
+	rules, valid, invalid := s.FetchVexFromUpstream(ctx, asset.ID, externalURLs)
+
+	if err := s.externalReferenceRepository.SaveBatch(ctx, tx, append(valid, invalid...)); err != nil {
```
Hubtrick-Git, line 999: I think we should return the error here if the save batch fails

services/fetch_service.go
```diff
@@ -79,92 +78,10 @@ func VexRulesFromDocument(body []byte, source string) ([]models.UpstreamVEXRule,
...
-	}
-
-	return boms, validURLs, invalidURLs
-}
-
 // FetchVexFromUpstream downloads VEX from the given external references and converts it into
 // VEX rules scoped to the given asset version. CSAF, OpenVEX and CycloneDX are parsed to rules by
 // their respective transformers; the returned rules carry their VexSource (the reference URL).
-func FetchVexFromUpstream(ctx context.Context, assetID uuid.UUID, string, upstreamURLs []string) ([]models.VEXRule, []models.ExternalReference, []models.ExternalReference) {
+func FetchVexFromUpstream(ctx context.Context, assetID uuid.UUID, upstreamURLs []string) ([]models.VEXRule, []models.ExternalReference, []models.ExternalReference) {
```
Hubtrick-Git: This function is also currently dead duplicate code and safe to delete

Hubtrick-Git (changes requested):

services/scan_service.go
Hubtrick-Git, line 642: do we want to handle the "else" block of this as well?

Hubtrick-Git:
Fixed some bugs unrelated to the PR which I found while reviewing the code

Hubtrick-Git (approved):
Good job !

timbastin (approved):

timbastin:
The only thing i dislike in this pr is that it keep making our callstack deeper: 

Update artifact -> fetch sboms -> ingest vex -> save...

It might be hard to reuse those functions later on. Nevertheless, the solution is fine and we need to think about a refactoring later on

PR #3023, opened 2026-09-06
Feat/SBOM merkle tree

timbastin:


database/repositories/sbom_repository.go
```diff
@@ -0,0 +1,234 @@
...
+//
+// Content addressing means an edge is never updated in place and never deleted
+// on rescan, so unreferenced subtrees accumulate: every superseded revision of
+// every SBOM leaves its old spine behind. Only reachability from the sboms
+// table can tell those apart from subtrees still shared by someone else, which
+// is why this is a mark and sweep rather than reference counting.
+//
+// It returns the number of deleted edges.
+func (r *sbomRepository) CollectGarbage(ctx context.Context, tx *gorm.DB) (int64, error) {
+	result := r.GetDB(ctx, tx).Exec(`
```
Hubtrick-Git, line 220: will this function ever realistically terminate? This has to be a computational explosion, no?

database/repositories/sbom_repository.go
```diff
@@ -0,0 +1,234 @@
...
+//
+// Edges are inserted with ON CONFLICT DO NOTHING, so subtrees already stored -
+// by this artifact, another artifact, or another organization entirely - cost
+// nothing. That is where the storage saving comes from.
+//
+// Re-ingesting a source moves its row to the new root hash instead of adding
+// one per content revision. The superseded root's subtrees stay for the garbage
+// collector, since other SBOMs may still reference them.
+func (r *sbomRepository) SaveTree(ctx context.Context, tx *gorm.DB, sbom models.SBOM, tree *normalize.MerkleTree) error {
+	db := r.GetDB(ctx, tx)
```
Hubtrick-Git, line 61: Maybe we should guard this inside a transaction/rollback statement, if the caller did not provide one themselves. There are a lot of call sites and its probably safer not to trust them with transaction control.

database/migrations/20260905120000_sbom_merkle_tree.up.sql
```diff
@@ -0,0 +1,77 @@
...
+    PRIMARY KEY (root_subtree_hash, artifact_name, asset_version_name, asset_id, source),
+    CONSTRAINT fk_sboms_asset_version
+        FOREIGN KEY (asset_version_name, asset_id)
+        REFERENCES public.asset_versions (name, asset_id) ON DELETE CASCADE
+);
+
+CREATE INDEX IF NOT EXISTS idx_sboms_asset_version
+    ON public.sboms (asset_id, asset_version_name);
+
+CREATE INDEX IF NOT EXISTS idx_sboms_root
```
Hubtrick-Git: this is already covered by the primary key index

database/migrations/20260905120000_sbom_merkle_tree.up.sql
```diff
@@ -0,0 +1,77 @@
...
+-- non-unique by design (a component with n children has n rows), so it cannot
+-- be a foreign key target.
+CREATE TABLE IF NOT EXISTS public.sboms (
+    root_subtree_hash  UUID NOT NULL,
+    artifact_name      TEXT NOT NULL,
+    asset_version_name TEXT NOT NULL,
+    asset_id           UUID NOT NULL,
+    source             TEXT NOT NULL,
+    updated_at         TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (root_subtree_hash, artifact_name, asset_version_name, asset_id, source),
```
Hubtrick-Git: Regarding the index structure on this table: I think we should rearrange the primary key to lead with asset_Id, asset_version_name . That way we can drop the idx_sboms_asset_version Index . Then we could also debate to drop the root_subtree_hash index but that is probably unsignificant on such a small table

database/migrations/20260905120000_sbom_merkle_tree.up.sql
```diff
@@ -0,0 +1,77 @@
...
+    -- it, so two component ids under one subtree hash would be a sha256
+    -- collision. Including it only widened the index by a purl per row.
+    -- NULLS NOT DISTINCT so a leaf row can only be inserted once (PG15+).
+    CONSTRAINT sbom_merkle_edges_unique
+        UNIQUE NULLS NOT DISTINCT (subtree_hash, direct_dependency_subtree_hash)
+);
+
+
+-- upward traversal (vulnerable purl -> affected SBOMs)
+CREATE INDEX IF NOT EXISTS idx_sbom_merkle_edges_child
```
Hubtrick-Git, line 47: Regarding this index, we could also add subtree_hash to it, so we can possible do index-only scans for example in the FindSBOMsContainingComponent query.

cmd/devguard-cli/hashmigrations/merkle_backfill.go
```diff
@@ -0,0 +1,250 @@
...
+
+	slog.Info("backfilling merkle sboms from component_dependencies", "assetVersions", len(assetVersions))
+
+	sbomRepository := repositories.NewSBOMRepository(db)
+
+	// One transaction for the whole migration, including the drop: a partial
+	// backfill followed by a dropped source table would be unrecoverable, so
+	// either every asset version is migrated and the table goes, or nothing
+	// changes at all.
+	return db.Transaction(func(tx *gorm.DB) error {
```
Hubtrick-Git, line 210: If this migration is taking very long we could definitely improve the performance of this, for example dropping the indexes and constraints before inserting and then building them afterwards.

Hubtrick-Git (commented):
Some comments regarding some database aspects of this pull request

database/repositories/sbom_repository.go
timbastin: did that already :D

database/migrations/20260905120000_sbom_merkle_tree.up.sql
timbastin: Wow i did not even know that about postgresql b-trees. Will do!

database/migrations/20260905120000_sbom_merkle_tree.up.sql
timbastin: Wow great!

database/repositories/sbom_repository.go
timbastin: It just takes 1 second currently. Seems pretty fast to travers the tree... I do not have deeper knowledge about why this is so fast...

cmd/devguard-cli/hashmigrations/merkle_backfill.go
timbastin: I think its fast enough currently. Takes around 10 seconds.

PR #3029, opened 2026-09-09
Implemented error log storage

Dboy0ZDev:
Related Issue: https://github.com/l3montree-dev/devguard/issues/1159

This PR does not directly solve https://github.com/l3montree-dev/devguard/issues/1159. It implements error logging storage as a related idea to this issue.

Webhook issues are not monitored yet and therefore not stored.

### DIFF
--- a/controllers/log_controller.go
+++ b/controllers/log_controller.go
@@ -26,7 +26,7 @@ func NewLogController(logService shared.LogService) *LogController {
 // @Param pageSize query int false "Page size"
 // @Success 200 {object} shared.Paged[models.Log]
 // @Router /organizations/{organization}/projects/{projectSlug}/assets/{assetSlug}/logs/ [get]
-func (controller LogController) ListPaged(ctx shared.Context) error {
+func (controller *LogController) ListPaged(ctx shared.Context) error {
 	org := shared.GetOrg(ctx)
 	project := shared.GetProject(ctx)
 	asset := shared.GetAsset(ctx)
--- a/controllers/log_controller.go
+++ b/controllers/log_controller.go
@@ -1,0 +1,37 @@
+package controllers
+
+import (
+	"github.com/l3montree-dev/devguard/database/models"
+	"github.com/l3montree-dev/devguard/shared"
+	"github.com/labstack/echo/v4"
+)
+
+type LogController struct {
+	logService shared.LogService
+}
+
+func NewLogController(logService shared.LogService) *LogController {
+	return &LogController{logService: logService}
+}
+
+// @Summary List logs
+// @Tags Logs
+// @Security CookieAuth
+// @Security PATAuth
+// @Security BearerAuth
+// @Param organization path string true "Organization slug"
+// @Param projectSlug path string true "Project slug"
+// @Param assetSlug path string true "Asset slug"
+// @Success 200 {object} shared.Paged[models.Log]
+// @Router /organizations/{organization}/projects/{projectSlug}/assets/{assetSlug}/logs [get]
+func (controller LogController) ListPaged(ctx shared.Context) error {
+	org := shared.GetOrg(ctx)
+	project := shared.GetProject(ctx)
+	asset := shared.GetAsset(ctx)
+	var logs shared.Paged[models.Log]
+	logs, err := controller.logService.ListPaged(ctx, nil, org.ID, project.ID, asset.ID)
+	if err != nil {
+		return echo.NewHTTPError(500, "could not get logs").WithInternal(err)
+	}
+	return ctx.JSON(200, logs)
+}
--- a/controllers/log_controller.go
+++ b/controllers/log_controller.go
@@ -22,8 +22,10 @@ func NewLogController(logService shared.LogService) *LogController {
 // @Param organization path string true "Organization slug"
 // @Param projectSlug path string true "Project slug"
 // @Param assetSlug path string true "Asset slug"
+// @Param page query int false "Page number"
+// @Param pageSize query int false "Page size"
 // @Success 200 {object} shared.Paged[models.Log]
-// @Router /organizations/{organization}/projects/{projectSlug}/assets/{assetSlug}/logs [get]
+// @Router /organizations/{organization}/projects/{projectSlug}/assets/{assetSlug}/logs/ [get]
 func (controller LogController) ListPaged(ctx shared.Context) error {
 	org := shared.GetOrg(ctx)
 	project := shared.GetProject(ctx)
--- a/database/migrations/20260909072532_add_log_model.up.sql
+++ b/database/migrations/20260909072532_add_log_model.up.sql
@@ -4,7 +4,7 @@ CREATE TABLE IF NOT EXISTS public.logs (
     project_id uuid,
     asset_id uuid,
     asset_version_name text,
-    created_at timestamp NOT NULL DEFAULT now(),
+    created_at timestamp with time zone NOT NULL DEFAULT now(),
     log_level int NOT NULL,
     message text NOT NULL
 );
--- a/database/migrations/20260909072532_add_log_model.up.sql
+++ b/database/migrations/20260909072532_add_log_model.up.sql
@@ -1,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS public.logs (
+    id uuid NOT NULL PRIMARY KEY DEFAULT gen_random_uuid(),
+    org_id uuid NOT NULL,
+    project_id uuid,
+    asset_id uuid,
+    asset_version_name text,
+    created_at timestamp NOT NULL DEFAULT now(),
+    log_level int NOT NULL,
+    message text NOT NULL
+);
--- a/database/migrations/20260909072604_drop_pipelineerror_asset.up.sql
+++ b/database/migrations/20260909072604_drop_pipelineerror_asset.up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE public.assets DROP COLUMN IF EXISTS pipeline_error;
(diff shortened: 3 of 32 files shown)


controllers/log_controller.go
```diff
@@ -0,0 +1,39 @@
...
+// @Param assetSlug path string true "Asset slug"
+// @Param page query int false "Page number"
+// @Param pageSize query int false "Page size"
+// @Success 200 {object} shared.Paged[models.Log]
+// @Router /organizations/{organization}/projects/{projectSlug}/assets/{assetSlug}/logs/ [get]
+func (controller *LogController) ListPaged(ctx shared.Context) error {
+	org := shared.GetOrg(ctx)
+	project := shared.GetProject(ctx)
+	asset := shared.GetAsset(ctx)
+	var logs shared.Paged[models.Log]
```
Hubtrick-Git, line 33: ineffectual declare: ":=" operator redeclares all variables on the left side

database/migrations/20260909072604_drop_pipelineerror_asset.up.sql
Hubtrick-Git, line 1: maybe merge both migrations into 1 file to minimize the amount of migration files inside the migrations folder?

services/log_service.go
```diff
@@ -0,0 +1,130 @@
...
+	return &logService{
+		logRepository:          logRepository,
+		organizationRepository: organizationRepository,
+		projectRepository:      projectRepository,
+		assetRepository:        assetRepository,
+		assetVersionRepository: assetVersionRepository,
+	}
+}
+
+func (s logService) cascadeIDs(ctx context.Context, tx *gorm.DB, orgID, projectID, assetID uuid.UUID, assetVersionName string) (uuid.UUID, uuid.UUID, uuid.UUID, string, error) {
```
Hubtrick-Git, line 38: This function definitely needs a few comments to explain whats happening and why we need that

services/log_service.go
```diff
@@ -0,0 +1,130 @@
...
+
+	err = s.logRepository.Save(ctx, tx, &log)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s logService) StoreRecoverPanic(ctx context.Context, tx *gorm.DB, orgID uuid.UUID, projectID uuid.UUID, assetID uuid.UUID, assetVersionName string, message string) error {
```
Hubtrick-Git, line 94: This function and StoreCaptureException are identical except the name. I would remove 1 and generalize the name

monitoring/alert.go
```diff
@@ -16,32 +16,95 @@
...
 	"github.com/pkg/errors"
+	"gorm.io/gorm"
 )
 
-func Alert(message string, err error) {
+type LogService interface {
+	StoreCaptureException(ctx context.Context, tx *gorm.DB, orgID uuid.UUID, projectID uuid.UUID, assetID uuid.UUID, assetVersionName string, message string) error
+	StoreRecoverPanic(ctx context.Context, tx *gorm.DB, orgID uuid.UUID, projectID uuid.UUID, assetID uuid.UUID, assetVersionName string, message string) error
+}
+
```
Hubtrick-Git, line 34: We have all these options and a lot of code to cascade and retrieve scoped logs, yet **we never even set the scope or the tx once**. Only ctx gets passed making all the other attributes/columns *currently* useless.

database/migrations/20260909072532_add_log_model.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS public.logs (
+    id uuid NOT NULL PRIMARY KEY DEFAULT gen_random_uuid(),
+    org_id uuid NOT NULL,
+    project_id uuid,
```
Hubtrick-Git, line 4: I think the data layout is a bit confusing, for project ids and asset ids we allow NULL but not for orgID, but  this is never enforced in the code. I think it only works because you always pass uuid.Nil everywhere instead of uuid pointers (those being nilable). This leads to a lot of confusion and also if just minor takes up unnecessary space in the table. 

Also I think it would be more readable to use a textual representation for the log level. Way easier to handle the data this way and people who do not remember don't waste time looking up the different log levels.

monitoring/alert.go
```diff
@@ -16,32 +16,95 @@
...
+		storedMsg = fmt.Sprintf("%s: %v", message, err)
+	}
+	loggerErr := logger.StoreCaptureException(ctx, opts.Tx, opts.OrgID, opts.ProjectID, opts.AssetID, opts.AssetVersionName, storedMsg)
+	if loggerErr != nil {
+		slog.Error("could not store error in database", "msg", message, "err", loggerErr)
+	}
 }
 
-func RecoverPanic(msg string) {
+func RecoverPanic(msg string, opts AlertOptions) {
```
Hubtrick-Git, line 94: Previously this function was designed as a no op if the caller did not panic. Only if it paniced it called 1 alert. Now it calls an alert every time AND even logs 2 alerts on panic. Revert back to only 1 call on panic.

monitoring/alert.go
```diff
@@ -16,32 +16,95 @@
...
 	}
+	ctx := opts.Ctx
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if logger == nil {
+		slog.Error("could not store error in database", "msg", "logger has not been set yet")
+		return
+	}
+	loggerErr := logger.StoreCaptureException(ctx, opts.Tx, opts.OrgID, opts.ProjectID, opts.AssetID, opts.AssetVersionName, message)
```
Hubtrick-Git, line 66: I think this functions calls on every gorm error. Inside the function we eventually call a few gorm functions. In theory if an error would occur inside one of these gorm functions itself we would indefinitely call the db until something crashes. Maybe worth to take a look at

Hubtrick-Git (changes requested):
There are some major bugs inside the code which need to be adressed beforehand.

A few comments also target the data layout of your logs table and the go functions handling those logs. Maybe have a look at that if we maybe change it up a bit.

Last but not least a few minor corrections regarding style and code best practices.

timbastin:
Lets go with two distinct functions: 

monitoring.Alert
monitoring.AlertAndSaveInErrorLog(ctx, options, message, err)

We should show error logs without org id in the instance admin dashboard.

Lets go with 200 error logs

timbastin:
After this is finished. We should follow up deleting the content tree. https://github.com/l3montree-dev/devguard/issues/2740

controllers/org_controller.go
```diff
@@ -647,3 +651,77 @@ func (controller *OrgController) RevokeInvitation(ctx shared.Context) error {
...
+	assetID := ctx.QueryParam("assetid")
+
+	switch {
+	case assetID != "":
+		id, err := uuid.Parse(assetID)
+		if err != nil {
+			return echo.NewHTTPError(400, "could not parse asset id").WithInternal(err)
+		}
+
+		organizationSlug, projectSlug, assetSlug, err := controller.assetRepository.GetOrgProjectAssetSlugsByAssetID(reqCtx, nil, id)
```
timbastin, line 704: We need to check if the user has access to this asset, project and org. Otherwise we might leak some information we don't want to

database/repositories/asset_repository.go
```diff
@@ -68,6 +69,19 @@ func (repository *assetRepository) prepareUniqueSlugs(ctx context.Context, tx *g
 	return nil
 }
 
+func (repository *assetRepository) ReadWithoutErrorLog(ctx context.Context, tx *gorm.DB, id uuid.UUID) (models.Asset, error) {
+	var asset models.Asset
+	db := repository.GetDB(ctx, tx).Session(&gorm.Session{
+		Logger:               logger.Default.LogMode(logger.Silent),
+		FullSaveAssociations: false,
```
timbastin, line 76: I don't really get why we need this line.

database/repositories/asset_version_repository.go
```diff
@@ -59,6 +60,15 @@ func (repository *assetVersionRepository) Read(ctx context.Context, tx *gorm.DB,
 	return asset, err
 }
 
+func (repository *assetVersionRepository) ReadWithoutErrorLog(ctx context.Context, tx *gorm.DB, assetVersionName string, assetID uuid.UUID) (models.AssetVersion, error) {
+	var asset models.AssetVersion
+	err := repository.GetDB(ctx, tx).Session(&gorm.Session{
+		Logger:               logger.Default.LogMode(logger.Silent),
+		FullSaveAssociations: false,
```
timbastin, line 67: Same here

database/repositories/project_repository.go
```diff
@@ -29,6 +30,15 @@ func NewProjectRepository(db *gorm.DB) *projectRepository {
 	}
 }
 
+func (g *projectRepository) ReadWithoutErrorLog(ctx context.Context, tx *gorm.DB, id uuid.UUID) (models.Project, error) {
+	var result models.Project
+	err := g.GetDB(ctx, tx).Session(&gorm.Session{
+		Logger:               logger.Default.LogMode(logger.Silent),
+		FullSaveAssociations: false,
+	}).Model(models.Project{}).Where("ID = ?", id).First(&result).Error
```
timbastin: Uppercase id right here?

timbastin (changes requested):

database/repositories/asset_repository.go
Dboy0ZDev: This was a change in response to something @Hubtrick-Git mentioned correctly.

`Alert()` calls `SaveLog()` which internally calls `cascadeIDs` which contains a Database read. If the database fails, it triggers another log (since log-on-error seems to be the default config). We would then return back to the beginning of the log cycle and start another alert. We risk running into an infinite loop.

Therefore `cascadeIds()` required a silent read. If that read fails, we don't run into an infinite loop.

PR #3035, opened 2026-09-10
Fixed asset overview calculation

### LINKED ISSUE #3016: Asset overview CVE count does not match after findings where handled
Overview reports critical CVEs per CVSS, those were closed and are still shown on overview.

Image

Image

Dboy0ZDev:
Issue: https://github.com/l3montree-dev/devguard/issues/3016

- Added a Preload of "Artifact" when deleting a vexrule to trigger vuln event patching
- Added groupEvent fetch to consider all events when vexrule is used to update Rsik assessment

### DIFF
--- a/database/repositories/dependency_vuln_repository.go
+++ b/database/repositories/dependency_vuln_repository.go
@@ -6,6 +6,7 @@ import (
 	"iter"
 	"log/slog"
 	"sort"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/l3montree-dev/devguard/dtos"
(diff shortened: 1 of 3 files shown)


database/repositories/dependency_vuln_repository.go
```diff
@@ -78,17 +79,17 @@ func (repository *dependencyVulnRepository) GetByVexRuleID(ctx context.Context,
...
 		return nil, err
 	}
-	if err := repository.attachGroupEvents(ctx, tx, dependencyVulns); err != nil {
+	if err := AttachGroupEvents(repository.GetDB(ctx, tx), dependencyVulns, nil); err != nil {
 		return nil, err
 	}
 	return dependencyVulns, nil
 }
 
-func (repository *dependencyVulnRepository) attachGroupEvents(ctx context.Context, tx *gorm.DB, vulns []models.DependencyVuln) error {
```
Hubtrick-Git, line 91: why did you change it from a method to a function? seems unintuitive

Dboy0ZDev: This is actually a remnant from previous solutions I tried that required the function to be global. The final solution does not require it anymore, but I forgot to revert that addition. Will change it back.

Dboy0ZDev: This is actually a remnant from previous solutions I tried that required the function to be global. The final solution does not require it anymore, but I forgot to revert that addition. Will change it back.

Hubtrick-Git (approved):

timbastin:
I am not sure if the performance might degrade since now there are multiple sql statements in the hot path of travel dependency vuln state, but we can try to go with this.

You have a lot of changes in the mocks package. The reason for this is that you are using another mockery version for generating the mocks. Please try todo the following

`nix develop -c make mocks`

This will use the mockery version defined in the nix flake, which should be byte identical for all of us.

